PR #1, opened 2015-07-20
Add exception handling and fix minor issues

nghiant2710:


application.py
```diff
@@ -9,4 +9,15 @@ def restart(self, app_id):
         return self.request('/application/' + str(app_id) + '/restart', 'POST')
 
     def get_by_id(self, app_id):
-        return self.request('application?id=' + str(app_id), 'GET')[0]
+    	params = {
```
dfunckt: The whole block is badly indented here

dfunckt:
Why so many print statements?

Btw, prefer to use the print() function instead, it's python 3 compatible.

resources.py
```diff
@@ -0,0 +1,6 @@
+#
+#    This is a python library for resources like message templates etc.
+#
+
+class Message:
```
dfunckt: It's best to always extend from object, ie:

Class Message(object):
  # ...

Old style classes are deprecated and I think don't even work on Python 2.7

device.py
```diff
@@ -27,14 +36,21 @@ def get_by_name(self, name):
...
+            # unexpected exception
+            pass
 
     def identify(self, uuid):
         data = {
             'uuid': uuid
         }
         return self.request('/blink', 'POST', data=data)
-
+
```
dfunckt: Stray whitespace here.

device.py
```diff
@@ -27,14 +36,21 @@ def get_by_name(self, name):
...
         }
-        return self.request('device', 'GET', params=params)[0]
+        try:
+            return self.request('device', 'GET', params=params)
+        except IndexError:
+            # found no device
+            print Message.NO_DEVICE_FOUND.format(value=uuid,dev_att="name")
+        except:
+            # unexpected exception
+            pass
```
dfunckt: Shouldn't this be somehow handled here?

device.py
```diff
@@ -1,4 +1,6 @@
 from .baseapi import BaseAPI
+from resources import Message
+import sys
```
dfunckt: I personally find sorting my imports sensibly helps reviewing my code.

I follow this rules:

- First python imports
- then third party libraries
- then project/package imports
- then local imports

All "sections" are kept sorted by name and separated by a blank line. I also put relative imports below absolute ones.

dfunckt:
I find some methods are not consistent in returning things -- you should make sure return values make sense and are not a dict, or an int or None, etc, depending on conditions.

nghiant2710:
@dfunckt I've updated your comments.
About returning things, some methods are not required a return value, like API call to blink a device. So I just leave it with a return to break from that function.

device.py
```diff
@@ -1,5 +1,7 @@
-from .baseapi import BaseAPI
+import sys
 
+from .baseapi import BaseAPI
+from .resources import Message
```
dfunckt: Ah, that's better :)

resources.py
```diff
@@ -0,0 +1,7 @@
+#
+#    This is a python library for resources like message templates etc.
+#
```
dfunckt: Minor thing: prefer to document Python code in triple-quote blocks. For a module "docstring" I would write it to the top like this:

```
  """
  This module is doing this and that.
  """
```

Similarily for function, class or method, I'd codument it just below the signature, ie:

```
class Foo(object):
  """
  Foo blah..
  """
  def bar(self):
    """
    This is a method docstring
    """
```

Python (and third-party tools) does a great job taking those "docstrings" and generating documentation.

resources.py
```diff
@@ -0,0 +1,7 @@
...
+
+class Message(object):
+	NO_DEVICE_FOUND="There is no device with {dev_att}: {value}! Please check the device {dev_att}!"
+	NO_APPLICATION_FOUND="There is no application with {app_att}: {value}! Please check the application {app_att}!"
```
dfunckt: Also minor, there should be a space around the assignment "=" there.

dfunckt:
Looks good overall :) If you like, please address a couple of new comments I left, but they're minor things considering this is a work in progress.

Re: return values, my comment may not apply to this PR, given that I read it out of context (ie. this diff). It still holds though as a general suggestion/best practice.

:)

application.py
```diff
@@ -9,4 +9,15 @@ def restart(self, app_id):
...
+            'eq': app_id
+        }
+        try:
+            return self.request('application', 'GET', params=params)
+        except IndexError:
+            # found no application
+            print(Message.NO_APPLICATION_FOUND.format(value=app_id,dev_att="id"))
+        except:
+            # unexpected exception
+            pass
```
dfunckt: You still don't handle the exceptions -- same in ``device.py`` -- you just silence them. I think users of the library would like to know that an error has happened.

nghiant2710: @dfunckt . My bad. I didn't notice that. I was testing all the methods so wanted to ignore unexpected errors.

PR #2, opened 2015-07-30
Device, Application and Config added!

nghiant2710:


application.py
```diff
@@ -1,23 +1,81 @@
...
+            return False
 
-    def restart(self, app_id):
-        return self.request('/application/' + str(app_id) + '/restart', 'POST')
+    def has_any(self):
+        apps = self.request('application', 'GET')['d']
+        if apps:
+            return True
+        else:
+            return False
```
dfunckt: You could write this ``if`` block as: ``return bool(apps)``

application.py
```diff
@@ -1,23 +1,81 @@
...
-            # unexpected exception
-            raise
+            print(Message.NO_APPLICATION_FOUND.format(value=app_id,app_att="id"))
+
+    def create(self, name, device_type):
+        device_slug = self.device.get_device_slug(device_type)
+        if device_slug:
+            data = {
+                'app_name': name,
+                'device_type': device_slug[0]
```
dfunckt: I would expect that ``self.device.get_device_slug(device_type)`` would give me a string. Why not?

application.py
```diff
@@ -1,23 +1,81 @@
...
+        params = {
+            'filter': 'app_name',
+            'eq': name
+        }
+        return self.request('application', 'DELETE', params=params)            
+
+    def restart(self, name):
+        app = self.get(name)
+        if app:
+            return self.request('/application/' + str(app[0]['id']) + '/restart', 'POST')
```
dfunckt: It's better not to concatenate strings this way. Better use ``format``: ``self.request('/application/{0}'.format(app[0]['id']))``.

baseapi.py
```diff
@@ -115,4 +130,4 @@ def request(self, url, method, params=None, data=None):
         except ValueError:
             raise JSONDecodeError()   
 
-        return json['d']
+        return json
```
dfunckt: What keys does ``json`` contain? What was this ``['d']`` key that was returned before?

device.py
```diff
@@ -20,23 +28,67 @@ def get(self, uuid):
...
         params = {
-            'expand': 'application',
-            'eq': name,
-            'filter': "app_name"
+            'filter': 'app_name',
+            'eq': name
         }
-        return self.request('device', 'GET', params=params)
+        app = self.request('application', 'GET', params=params)['d']
+        print app
```
dfunckt: Whoops :p

device.py
```diff
@@ -20,23 +28,67 @@ def get(self, uuid):
...
+        return self.get(uuid)['is_online']
+
+    def get_local_ip_address(self, uuid):
+        if self.is_online(uuid):
+            device = self.get(uuid)
+            ips = device['ip_address'].split()
+            ips.remove(device['vpn_address'])
+            return ips
+        else:
+            print(Message.DEVICE_OFFLINE.format(uuid=uuid))
```
dfunckt: This method is called ``get_local_ip_address()`` and returns either a list of somethings or nothing (but prints a message to the console). Imagine how many different cases you'd have to handle if you were a user of the SDK and all you wanted was to get your device's IP.

device.py
```diff
@@ -44,8 +96,72 @@ def identify(self, uuid):
...
+            params = {
+                'filter': 'uuid',
+                'eq': uuid
+            }
+            data = {
+                'name': new_name
+            }
+            return self.request('device', 'PATCH', params=params, data=data)
+        else:
+            print(Message.NO_DEVICE_FOUND.format(value=uuid,dev_att="uuid"))
```
dfunckt: Again, this method is not clear what it returns, or it returns many different things depending on circumstances. We should generally be consistent with how we handle API calls and what values we return so that a user of our library doesn't have to worry and mess with ``if this: blah() else: blablah()``

device.py
```diff
@@ -44,8 +96,72 @@ def identify(self, uuid):
...
+
+    def get_display_name(self, device_type_slug):
+        device_type_found = self.config.get_device_type()
+        display_name = [device['name'] for device in device_type_found
+                        if device['slug'] == device_type_slug]
+        if display_name:
+            return display_name
```
dfunckt: Seeing the list comprehension you do above (lines 128-129), ``display_name`` is a list!

dfunckt:
Great work Trong! My comments are mostly about minor style fixes, but there is one issue I'd like to insist on, and that is how we handle API responses and what things we return to the user of the SDK.

We must always think how a user will use our library and how to make things convenient to them -- why bother with an intermediate layer of code (i.e. the SDK) if not to simplify things? Doing this is difficult and we get better as we gain experience. My suggestion is that you try to review what values each method returns for each case it handles, and whether the returned value would be expected by the user.

nghiant2710:
@dfunckt , I argee with you, I think we should have a list of exitcode and message for user convenience. I plan to discuss with Juanchi later about this. At first, just want make this works like the JS sdk then polishes everything at the end

application.py
```diff
@@ -1,23 +1,81 @@
...
+        app = self.request('application', 'GET', params=params)['d']
+        if app:
+            return app
+        else:
             # found no application
-            print(Message.NO_APPLICATION_FOUND.format(value=app_id,dev_att="id"))
-        except:
-            # unexpected exception
-            raise
+            print(Message.NO_APPLICATION_FOUND.format(value=app_id,app_att="id"))
```
emirotin: space after comma :)

baseapi.py
nghiant2710: ['d'] represents for response data. In some cases, it's not json obj, just "OK" so I remove from here and handle the data in each method

device.py
nghiant2710: As I comment below, this will need to be polished later

__init__.py
```diff
@@ -12,5 +14,5 @@ def __init__(self, token=None):
         self.application = Application(token)
         self.device = Device(token)
         self.environment_variables = Environment_Variables(token)
-
-
+        self.config = Config(token)
+    	self.key = Key(token)
```
dfunckt: wrong indentation :)

PR #5, opened 2015-08-11
New version of resin python sdk

nghiant2710:
This PR changes python sdk structure the same as resin-sdk 
- Authentication, settings, token, os, key added!
- Remake baseapi -> base_request
- Move application, device, config, os, key, env variable to Models packages


auth.py
```diff
@@ -0,0 +1,49 @@
...
+		return self.token.has()
+
+	def get_token(self):
+		return self.toket.get()
+
+	def get_user_id(self):
+		return self.token.get_user_id
+
+	def get_email(self):
+		return self.token.get_email
```
dfunckt: Shouldn't this be a method call? Same question for the one above, get_user_id

base_request.py
```diff
@@ -0,0 +1,149 @@
...
+    def __init__(self):
+        self.settings = Settings()
+        self.token = Token()
+        self.util = Util()
+
+    def __str__(self):
+        return b'<{:s} at {:#x}>'.format(type(self).__name__, id(self))
+
+    def __unicode__(self):
+        return '<{:s} at {:#x}>'.format(type(self).__name__, id(self))
```
dfunckt: I think you can do: str(self).encode('utf-8') to avoid repetition

base_request.py
```diff
@@ -0,0 +1,149 @@
...
+    def format_params(self, params):
+    	#TODO: make more effecient
+        if params:
+            if params.has_key('expand'):
+                query = "?$expand=" + \
+                        params[
+                            "expand"] + "($filter=" + params['filter'] + "%20eq%20'" + str(params["eq"]) + "')"
+            elif params.has_key('filter'):
+                query = "?$filter=" + params["filter"] + \
+                    "%20eq%20'" + str(params["eq"]) + "'"
```
dfunckt: This can be made a lot more readable with [string templates](https://docs.python.org/2/library/string.html#formatstrings).

settings.py
```diff
@@ -0,0 +1,84 @@
...
+CONFIG_FILENAME = 'resin.cfg'
+
+class Settings(object):
+
+	_setting = {
+		'pine_endpoint': 'https://api.resinstaging.io/ewa/',
+		'api_endpoint': 'https://api.resinstaging.io/',
+		'data_directory': Path.join(HOME_DIRECTORY, '.resin'),
+		# cache time : 1 week in milliseconds
+		'image_cache_time': (1 * 1000 * 60 * 60 * 24 * 7),
```
craig-mulligan: What is this setting for? Are you going to allow downloading of images?

README.md
```diff
@@ -5,17 +5,14 @@ A python wrapper for resin.io [work in progress]
 ## usage 
 
 ```
-pip install resin
```
craig-mulligan: I have the created the project on PyPI, to save the (name) But I will share with apo so you can deploy, when you think its ready for public.

README.md
```diff
@@ -5,17 +5,14 @@ A python wrapper for resin.io [work in progress]
...
-```
```
craig-mulligan: We should probably also mention that this is configured for resinstaging api incase any users come across this.

settings.py
nghiant2710: @craig-mulligan, downloading image actually works. You can find it in models/os.py. Sorry, I've created the documentation for this. the python sdk has the same method as JS sdk which are mentioned here https://github.com/resin-io/resin-sdk/blob/master/DOCUMENTATION.md. Only few methods (2 or 3 as I reckon) are not fully implemented yet. Please ask me if you need any supports

README.md
nghiant2710: Yeah, I removed this but we definitely will move it there soon

README.md
nghiant2710: @craig-mulligan , I've updated the README

base_request.py
nghiant2710: @dfunckt , this is from the old sources, I've not touched this. Anw, removed because It's not used anywhere

nghiant2710:
@dfunckt , @craig-mulligan, I've updated your comments

PR #6, opened 2015-08-19
Adding Logs module and Fix bugs

nghiant2710:
- Adding Logs module
- Adding exception handling.
- Rename Os module -> device_os
- Fix some bugs


auth.py
```diff
@@ -11,18 +12,18 @@ def __init__(self):
 		self.token = Token()
 
 	def login(self, **credentials):
-		try:
-			self.token.set(self.authenticate(**credentials))
-		except:
-			# TODO: need exception type
-			raise
+		token = self.authenticate(**credentials)
+		print token
```
emirotin: debug leftover?

exceptions.py
```diff
@@ -1,23 +1,74 @@
 # coding: utf-8
+from .resources import Message
 
-class APIException(Exception):
-    pass
+class ResinMissingOption(Exception):
```
dfunckt: No need to prepend these classes with "Resin"

resources.py
```diff
@@ -6,9 +6,17 @@ class Message(object):
...
+	RESIN_KEY_NOT_FOUND = "Key not found: {key}"
+	RESIN_DEVICE_NOT_FOUND = "Device not found: {device}"
+	RESIN_APPLICATION_NOT_FOUND = "Application not found: {application}"
+	RESIN_MALFORMED_TOKEN = "Malformed token: {token}"
+	RESIN_INVALID_DEVICE_TYPE = "Invalid device type: {dev_type}"
+	RESIN_INVALID_OPTION = "Invalid option: {option}"
+	RESIN_MISSING_OPTION = "Missing option: {option}"
+	RESIN_NON_ALLOWED_OPTION = "Non allowed option: {option}"
+	RESIN_LOGIN_FAILED = "Invalid credentials"
+	RESIN_DEVICE_OFFLINE = "Device is offline: {uuid}"
```
dfunckt: Same here, no need for the Resin prefix.

auth.py
```diff
@@ -2,6 +2,7 @@
 from .resources import Message
 from .token  import Token
 from .settings import Settings
+import exceptions as ResinException
```
dfunckt: This will explode in Python 3. There's no immediate need for Python 3 support, as other parts of the code also break on Python 3, but it's generally a good practice to do relative imports with a dot: "from . Import exceptions as ResinException"

base_request.py
```diff
@@ -10,12 +10,7 @@
 from .token  import Token
 from .resources import Message
 
-from exceptions import (
-    JSONDecodeError,
-    NoTokenProvided,
-    ResponseError,
-    RequestError
-)
+import exceptions as ResinException
```
dfunckt: Same here about relative imports

exceptions.py
```diff
@@ -1,23 +1,74 @@
...
+	def __init__(self, option):
+		self.code = 'ResinMissingOption'
+		self.exit_code = 1
+		self.message= Message.RESIN_MISSING_OPTION.format(option=option)
```
dfunckt: Needs space around "="

exceptions.py
```diff
@@ -1,23 +1,74 @@
 # coding: utf-8
+from .resources import Message
```
dfunckt: Why not import Message as ResinMessage similar to ResinException.

Personally I like more this import -- with the ResinBlah prefix -- but you should choose either one style and stick with it.

logs.py
```diff
@@ -1,9 +1,35 @@
...
+
 from .base_request import BaseRequest
 from .resources import Message
+from .models.config import Config
+
+class Logs(object):
+	def __init__(self):
+		self.config = Config()
+		
+
```
dfunckt: Two consecutive blank lines :)

logs.py
```diff
@@ -1,9 +1,35 @@
...
+	def subscribe(self, uuid, callback, error):
+		pubnub_key = self.config.get_pubnub_keys()
+		if not hasattr(self, 'pubnub'):
```
dfunckt: Why bother with hasattr -- just set self.pubnub in __init__ and check here for "if self.pubnub is None"

models/application.py
```diff
@@ -1,14 +1,15 @@
 from ..base_request import BaseRequest
 from ..resources import Message
-from .device import Device
 from ..settings import Settings
+from .config import Config
+import resin.exceptions as ResinException
```
dfunckt: Why do you need to alias the import? Just do "from .. import exceptions". Same for the other files.

models/application.py
```diff
@@ -41,21 +41,23 @@ def get_by_id(self, app_id):
...
-            # found no application
-            print(Message.NO_APPLICATION_FOUND.format(value=app_id, app_att="id"))
+        try:
+            return self.base_request.request('application', 'GET', params=params, endpoint=self.settings.get('pine_endpoint'))['d'][0]
+        except IndexError:
+            raise ResinException.ResinApplicationNotFound(app_id)
 
     def create(self, name, device_type):
-        device_slug = self.device.get_device_slug(device_type)
+        device_type_found = self.config.get_device_type()
```
dfunckt: Why get_device_type returns a list? It would save you the following list comprehension if it didn't.

base_request.py
```diff
@@ -93,35 +88,30 @@ def __request(self, url, method, params, endpoint, headers=None, data=None, stre
...
-        #print response.url
-        #
+
         if response.status_code == 201:
             return response.content
         # 204: no content
         if response.status_code == 204:
             return
+
+        if stream:
```
emirotin: shouldn't this go before the 201 status check?

models/environment_variables.py
```diff
@@ -91,6 +93,5 @@ def remove(self, var_id):
         }
         return self.base_request.request('environment_variable', 'DELETE', params=params, endpoint=self.settings.get('pine_endpoint'))
 
-    # TODO
-	#def is_system_variable(self, variable):
-
+    def is_system_variable(self, variable):
+        return bool(re.match('^RESIN_',variable))
```
dfunckt: You may just do "retutn variable.startswith('RESIN_')"

base_request.py
```diff
@@ -134,10 +124,12 @@ def request_new_token(self):
...
         response = requests.get(url, headers=headers)
+        if not response.ok:
+            raise ResinException.ResinRequestError(response._content)
         return response.content
 
 
 class Util(object):
 
     def should_update_token(self, age, token_fresh_interval):
-        return bool(age >= token_fresh_interval)
```
emirotin: I think conversion to bool is not needed here

```
>>> x = 3 > 5
>>> x
False
```

logs.py
```diff
@@ -1,9 +1,35 @@
...
+			self.pubnub = Pubnub(publish_key=pubnub_key['publish_key'], subscribe_key=pubnub_key['subscribe_key'])
+		channel = self.get_channel(uuid)
+		self.pubnub.subscribe(channels=channel, callback=callback, error=error)
+
+	def history(self, uuid, callback, error):
+		pubnub_key = self.config.get_pubnub_keys()
+		if not hasattr(self, 'pubnub'):
```
emirotin: this repetitive code (get key, check attr, init pubnub) can moved to a method. probably even a decorator, like `@needs_pubnub`?

also the call to `get_pubnub_keys` can be inside of `if`, right?

logs.py
```diff
@@ -1,9 +1,35 @@
...
+			self.pubnub = Pubnub(publish_key=pubnub_key['publish_key'], subscribe_key=pubnub_key['subscribe_key'])
+		channel = self.get_channel(uuid)
+		self.pubnub.history(channel=channel, callback=callback, error=error)
 
-#class Logs(object):
+	def unsubscribe(self, uuid):
+		if not hasattr(self, 'pubnub'):
```
emirotin: `not` seems to be here by mistake

logs.py
emirotin: BTW is `Pubnub` constructor an expensive operation? if not it can always be called from `__init__`

models/application.py
```diff
@@ -67,10 +69,10 @@ def remove(self, name):
     def restart(self, name):
         app = self.get(name)
```
emirotin: `get` switch to throw in case of error, so looks like no need for `if`?

instead if this method is supposed to be safe the `except` block is needed

(and the same below)

models/device.py
```diff
@@ -163,9 +169,22 @@ def generate_uuid(self):
...
+            'registered_at': (datetime.utcnow() - datetime.utcfromtimestamp(0)).total_seconds(),
+            'uuid': uuid
+        }
 
-    # NOT YET IMPLEMENTED, WAITING FOR AUTH
-    #def register(self, app_name, uuid):
+        if api_key:
+            data['apikey'] = api_key
 
+        return self.base_request.request('device', 'POST', data=data, endpoint=self.settings.get('pine_endpoint'))
```
emirotin: How is `pine_endpoint` different from `api_endpoint`? is it `api_endpoint + '/ewa'`? then maybe the config should better have `pine_prefix` instead? Or is it mirroring the model of Node SDK?

models/device_os.py
```diff
@@ -0,0 +1,66 @@
...
+
+NETWORK_WIFI = 'wifi'
+NETWORK_ETHERNET = 'ethernet'
+
+NETWORK_TYPES = [
+	NETWORK_WIFI,
+	NETWORK_ETHERNET
+]
+
+VALID_OPTIONS = [
```
emirotin: This module should ideally start using `device-type.json`. It can be fetched from the `API/config` endpoint

models/device_os.py
```diff
@@ -0,0 +1,66 @@
...
+
+class DeviceOs(object):
+
+	def __init__(self):
+		self.base_request = BaseRequest()
+		self.settings = Settings()		
+
+	def download(self, raw=None, **data):
+		self.params = self.parse_params(**data)
+		if raw:
```
emirotin: Can it be shortened?

```
response= self.base_request.request('download', 'POST', data=data, endpoint=self.settings.get('api_endpoint'), stream=True)
return raw and response.raw or response
```

models/device_os.py
```diff
@@ -0,0 +1,66 @@
...
+
+		if params['network'] not in NETWORK_TYPES:
+			raise ResinException.ResinInvalidOption('network')
+
+
+		if params['network'] == NETWORK_WIFI:
+			if 'wifiSsid' not in params:
+				raise ResinException.ResinMissingOption('wifiSsid')
+
+			if 'wifiKey' not in params:
```
emirotin: The key is optional AFAIK. At least it's not enforced in the UI

auth.py
nghiant2710: Removed

exceptions.py
nghiant2710: @dfunckt, I want to use the same class name with NodeJS sdk so our users won't be confused

resources.py
nghiant2710: In both SDKs, we usually use Resin prefix with messages and errors so I want to keep this in term of consistency

auth.py
nghiant2710: Updated

exceptions.py
nghiant2710: Changed Message to ResinMessage

logs.py
nghiant2710: I've changed pubnub init to a decorator as Eugene's suggestion

base_request.py
nghiant2710: @emirotin If there is no error, It will return 200 - OK. So before 200 status is fine

models/environment_variables.py
nghiant2710: Updated

logs.py
nghiant2710: @emirotin, Thanks for the suggestion. I've changed to a decorator. This really is an expensive operation, need to request keys which requires login and shouldn't be in `__init__`

logs.py
nghiant2710: updated

models/application.py
nghiant2710: Removed `if`. No need for `except` block here, `get` itself raises a correct exception

models/device.py
nghiant2710: Yeah, This SDK bases on model of Node SDK

models/device_os.py
nghiant2710: I will use device-types.json in the next PR

models/device_os.py
nghiant2710: updated

models/device_os.py
nghiant2710: updated

nghiant2710:
@emirotin, @dfunckt I've updated yours comments. Please have a look

base_request.py
emirotin: I don't understand. I've asked about the 20**1** status above. Appears that in that case you return `content` (string?) even if  `stream is True`

logs.py
```diff
@@ -1,9 +1,37 @@
...
 from .base_request import BaseRequest
-from .resources import Message
+from .models.config import Config
+
+class Logs(object):
+	def __init__(self):
+		self.config = Config()
+
+	def init_pubnub(func):
+		def wrapper(self, uuid, callback, error):
```
emirotin: This method shouldn't care about the wrapped function arguments, so can be `wrapper(*args, **kwargs)`?

models/device.py
```diff
@@ -121,39 +122,43 @@ def note(self, uuid, note):
             }
             return self.base_request.request('device', 'PATCH', params=params, data=data, endpoint=self.settings.get('pine_endpoint'))
         else:
-            print(Message.NO_DEVICE_FOUND.format(value=uuid, dev_att="uuid"))
+            raise ResinException.ResinDeviceNotFound(uuid)
 
     def get_display_name(self, device_type_slug):
-        device_type_found = self.config.get_device_type()
+        device_type_found = self.config.get_device_types()
```
emirotin: here and above the var name looks invalid, it reads as boolean, and has the "type" in singular. It's also only used once, so the method call can be in comprehension. BTW I think the `get_device_types` method chouls cache its results, and basically the Config class itself should only call the /config endpoint once because it's changing very seldom, so usually no need to re-fetch it throughout the app liefcycle

models/device_os.py
emirotin: Cool :)

resources.py
dfunckt: I understand, but Python is not Javascript -- Python has proper namespaces with no way to get (unintentionally at least) name clashes.

This is stylistic, but important: we should remove all ``RESIN_`` and ``Resin`` prefixes from classes and attributes -- it just doesn't make sense in Python. And also change our imports from ``from . import exceptions as ResinException`` to ``from . import exceptions`` and then just use it like ``raise exceptions.NotFoundError``, etc.

dfunckt: Same goes for ``ResinMessage`` -- it should just be ``Message``.

I know I rhetorically asked you "why not rename Message to ResinMessage similar to ResinException" but I intended for the opposite effect!

dfunckt: Here is a library written in a very pythonic way: [redis-python](https://github.com/andymccurdy/redis-py). This is the client for Redis, quite close in what we do here. Have a look at the code and try to find patterns and try to apply them to your code too.

logs.py
emirotin: Requesting config should not require login

logs.py
nghiant2710: Updated

models/device.py
nghiant2710: I've renamed the variable and cached `device_types` as well

resources.py
nghiant2710: @dfunckt I see your point, I've updated the classes and attributes name.

base_request.py
nghiant2710: request with stream=true only for downloading device Os and I only see the response with 200. Anw, I've updated as you suggest to make sure it works

nghiant2710:
@emirotin , @dfunckt I've updated the comments again. Please have a look :)

logs.py
nghiant2710: Yes, init config doesn't require login

dfunckt:
Looks good, Trong. Great work!

models/application.py
```diff
@@ -41,21 +40,23 @@ def get_by_id(self, app_id):
...
-            # found no application
-            print(Message.NO_APPLICATION_FOUND.format(value=app_id, app_att="id"))
+        try:
+            return self.base_request.request('application', 'GET', params=params, endpoint=self.settings.get('pine_endpoint'))['d'][0]
+        except IndexError:
+            raise exceptions.ApplicationNotFound(app_id)
 
     def create(self, name, device_type):
-        device_slug = self.device.get_device_slug(device_type)
+        device_type_found = self.config.get_device_types()
```
emirotin: Another place with obsolete variable name

models/config.py
```diff
@@ -16,5 +16,7 @@ def get_all(self):
 	def get_pubnub_keys(self):
 		return self.base_request.request('/config', 'GET', endpoint=self.settings.get('pine_endpoint'))['pubnub']
 
-	def get_device_type(self):
-		return self.base_request.request('/config', 'GET', endpoint=self.settings.get('pine_endpoint'))['deviceTypes']
+	def get_device_types(self):
```
emirotin: I meant that you can actually cache full response from `self.base_request.request('/config', 'GET', endpoint=self.settings.get('pine_endpoint'))` and consider it not changing for the entire app lifespan. Then individual methods would only return specific keys from there. This can be done with the decorator as well, like with pubnub initialization:

```
@load_config
def get_device_types(self):
  return self._config['deviceTypes']
```

models/environment_variables.py
```diff
@@ -91,6 +92,5 @@ def remove(self, var_id):
         }
         return self.base_request.request('environment_variable', 'DELETE', params=params, endpoint=self.settings.get('pine_endpoint'))
 
-    # TODO
-	#def is_system_variable(self, variable):
-
+    def is_system_variable(self, variable):
```
emirotin: there are also a couple of system vars: USER and RESIN

models/config.py
dfunckt: You could do it even better with a lazy property:

    @property
    def config(self):
      if not hasattr(self, '_config'):
        self._config = self.base_request.blah()
      return self._config

And you can use it like so:

    foo = self.config['bar']

models/config.py
```diff
@@ -9,12 +8,23 @@ class Config(object):
 	def __init__(self):
 		self.base_request = BaseRequest()
 		self.settings = Settings()
+		self._config = None
+
+	def _get_config(self, key):
+		if self._config:
+			if key in self._config:
```
emirotin: This check is not needed. If `_config` is defined the `get_all()` method will do nothing, so the `key` won't appear there anyway

PR #10, opened 2015-09-08
pep8 checker for PR and set supported python version to 2.x only

nghiant2710:


docs_generator.py
```diff
@@ -41,8 +41,7 @@ def print_functions(baseclass):
             baseclass, predicate=inspect.ismethod):
         if func_name is not '__init__' and not func_name.startswith('_'):
             func = getattr(baseclass, func_name)
-            print doc2md.doc2md(func.__doc__,
-                make_function_name(func, func_name), type=1)
+            print doc2md.doc2md(func.__doc__, make_function_name(func, func_name), type=1)
```
dfunckt: Let's use the `print()` function instead of the `print` expression -- it's forwards-compatible.

setup.py
```diff
@@ -48,10 +48,6 @@
         'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.6',
         'Programming Language :: Python :: 2.7',
-        'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.2',
-        'Programming Language :: Python :: 3.3',
-        'Programming Language :: Python :: 3.4',
```
dfunckt, line 54: Makes sense :)

tests/pep8-git-pr-checker
```diff
@@ -0,0 +1,46 @@
...
+            files.append(tmp_file)
+
+    tempdir = tempfile.mkdtemp()
+    for name in files:
+        filename = os.path.join(tempdir, name)
+        filepath = os.path.dirname(filename)
+        if not os.path.exists(filepath):
+            os.makedirs(filepath)
+        with file(filename, 'w') as f:
+            system('git', 'show', ':' + name, stdout=f)
```
dfunckt: Better use string format: `system('git', 'show', ':{0}'.format(name), stdout=f)`

tests/pep8-git-pr-checker
```diff
@@ -0,0 +1,46 @@
...
+
+    args = ['pep8']
+    args.extend(('--ignore', ','.join(ignore_codes)))
+    args.append('.')
+    output = system(*args, cwd=tempdir)
+    shutil.rmtree(tempdir)
+    if output:
+        print(output)
+        sys.exit(1)
```
dfunckt, line 42: Was the call to `exit` included in the `if` clause by accident? And why exit with status code `1` -- is there any error?

docs_generator.py
nghiant2710: ahh, my bad. Fixed!

tests/pep8-git-pr-checker
nghiant2710: updated

tests/pep8-git-pr-checker
nghiant2710, line 42: we use travis-ci to run pep8 check for each PR. if it detects any issues, exit is called to mark as the PR has failed pep8 check. Ex: https://travis-ci.org/resin-io/resin-sdk-python

nghiant2710, line 42: @dfunckt ^

dfunckt, line 42: Oh, yeah -- `if output` is not a boolean, it's like "if there's any output" -- makes sense now.

I thought it was passed in as an option, my bad.

docs_generator.py
```diff
@@ -55,40 +54,40 @@ def make_function_name(func, func_name):
 
 def main():
 
-    print doc2md.doc2md(resin.__doc__, 'Resin Python SDK', type=0)
+    print(doc2md.doc2md(resin.__doc__, 'Resin Python SDK', type=0))
     print_newline()
-    print ('## Table of Contents')
-    print doc2md.make_toc(Table_Of_Content)
+    print(('## Table of Contents'))
```
emirotin: Why double parens here?

nghiant2710: My bad. Removed!

PR #12, opened 2015-10-02
Resin SDK for Python v1.1.0

nghiant2710:
### Added
- Implement Resin API Key. User can authorize by credentials, auth token or set Resin API key as RESIN_API_KEY environment variable.
- Implement `resin.models.device.restart()`
- Implement `resin.models.device.has_device_url()`
- Implement `resin.models.device.get_device_url()`
- Implement `resin.models.device.enable_device_url()`
- Implement `resin.models.device.disable_device_url()`
### Changed
- Fix `resin.auth.is_logged_in()` remove auth token.
- Fix bug with API endpoints in some functions.
- Implement `login` flag that marks functions only work if users authorize by credentials or auth token.
### Removed
- Remove `VALID_OPTIONS` in resin.models.device_os that blocks extra os parameters when downloading an image.
- Remove `resin.models.config.get_pubnub_keys()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,26 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+This project adheres to [Semantic Versioning](http://semver.org/).
+
+## [1.1.0] - 2015-10-02
+
+### Added
+
+- Implement Resin API Key. User can authorize by credentials, auth token or set Resin API key as RESIN_API_KEY environment variable.
+- Implement `resin.models.device.restart()`
+- Implement `resin.models.device.has_device_url()`
+- Implement `resin.models.device.get_device_url()`
+- Implement `resin.models.device.enable_device_url()`
+- Implement `resin.models.device.disable_device_url()`
+
+### Changed
+
+- Fix `resin.auth.is_logged_in()` remove auth token.
+- Fix bug with API endpoints in some functions.
+- Implement `login` flag that marks functions only work if users authorize by credentials or auth token.
+
+### Removed
+
+- Remove `VALID_OPTIONS` in resin.models.device_os that blocks extra os parameters when downloading an image.
+- Remove `resin.models.config.get_pubnub_keys()`.
--- a/resin/exceptions.py
+++ b/resin/exceptions.py
@@ -199,6 +199,7 @@ class NotLoggedIn(Exception):
         self.exit_code = 1
         self.message = Message.NOT_LOGGED_IN
 
+
 class Unauthorized(Exception):
     """
     Exception when no user logged in and no Resin API Key provided.
@@ -251,6 +252,7 @@ class DeviceOffline(Exception):
         self.code = 'DeviceOffline'
         self.exit_code = 1
         self.message = Message.DEVICE_OFFLINE.format(uuid=uuid)
+
 
 class DeviceNotWebAccessible(Exception):
     """
--- a/resin/exceptions.py
+++ b/resin/exceptions.py
@@ -137,10 +137,10 @@ class DeviceNotFound(Exception):
 
     """
 
-    def __init__(self):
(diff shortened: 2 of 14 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,26 @@
...
+
+All notable changes to this project will be documented in this file.
+This project adheres to [Semantic Versioning](http://semver.org/).
+
+## [1.1.0] - 2015-10-02
+
+### Added
+
+- Implement Resin API Key. User can authorize by credentials, auth token or set Resin API key as RESIN_API_KEY environment variable.
+- Implement `resin.models.config.get_mixpanel_token()`.
```
dfunckt: Why is this public?

(Ignore my other comment on the actual commit)

resin/exceptions.py
```diff
@@ -137,7 +137,7 @@ class DeviceNotFound(Exception):
 
     """
 
-    def __init__(self):
+    def __init__(self, device):
```
dfunckt: I'd rename this to `uuid`. Also rename at `Messages.DEVICE_NOT_FOUND`

resin/exceptions.py
```diff
@@ -235,3 +252,23 @@ def __init__(self, uuid):
         self.code = 'DeviceOffline'
         self.exit_code = 1
         self.message = Message.DEVICE_OFFLINE.format(uuid=uuid)
+
+
+class DeviceNoWebAccessible(Exception):
```
dfunckt: DeviceNo**t**WebAccessible. Also the respective message constant.

resin/models/device.py
```diff
@@ -446,7 +446,7 @@ def register(self, app_name, uuid):
         application = self.application.get(app_name)
         api_key = self.base_request.request(
             '/application/{0}/generate-api-key'.format(application['id']),
```
dfunckt: I believe the leading `/` is not required?

resin/models/device.py
```diff
@@ -464,5 +464,110 @@ def register(self, app_name, uuid):
...
+
+        Args:
+            uuid (str): device uuid.
+
+        Raises:
+            DeviceNotFound: if device couldn't be found.
+
+        """
+
+        if not self.has(uuid):
```
dfunckt, line 531: I don't particularly like that this method is called `has`, it'd prefer `exists` or something. But anyway, not important.

setup.py
```diff
@@ -14,7 +14,7 @@
     # Versions should comply with PEP440.  For a discussion on single-sourcing
     # the version across setup.py and the project code, see
     # https://packaging.python.org/en/latest/single_source_version.html
-    version='1.0.3',
+    version='1.1.0',
```
dfunckt: I'd prefer we got the package version from the code itself (easiest way is method 6 at https://packaging.python.org/en/latest/single_source_version/#single-sourcing-the-project-version). For example:

```
# resin-sdk-python/resin/__init__.py

__version__ = '1.1.0'

# blah code...
```

And then in `setup.py`:

```
import resin

setup(
  ...
  version=resin.__version__,
  ...
)
```

resin/exceptions.py
nghiant2710: Updated!

resin/exceptions.py
nghiant2710: Updated!

resin/models/device.py
nghiant2710: Updated

resin/models/device.py
nghiant2710, line 531: I would like to keep that name because nodejs sdk uses it.

setup.py
nghiant2710: Updated

nghiant2710:
@dfunckt , @abresas . I've removed mixpanel and pubnub from the SDK. Please review it.

PR #13, opened 2015-10-15
fix setup.py install by parsing version string

forrestp:
setup.py was importing resin, which imports various dependencies that aren't installed yet. This fix parses the version string from **init**.py so that there is no need to import resin.


nghiant2710:
@forrestp, thanks for pointing this out. LGTM but can you see the convention check here https://travis-ci.org/resin-io/resin-sdk-python/jobs/85573076 and update the script?
Then It's good to go.

forrestp:
Done

PR #15, opened 2015-10-28
Implement 2fa support and resin supervisor

nghiant2710:


resin/models/supervisor.py
```diff
@@ -0,0 +1,363 @@
...
+
+        Returns:
+            str: `OK` signals that the supervisor is alive and well.
+
+        Raises:
+            InvalidOption: if the endpoint is Resin API proxy endpoint and device_id or app_id is not set.
+
+        """
+
+        if not self._on_device:
```
pcarranzav, line 105: It might be worth it to create a wrapper function so that you don't repeat this code for every endpoint?

resin/models/supervisor.py
```diff
@@ -0,0 +1,363 @@
...
+            )
+        else:
+            return self.base_request.request(
+                '{0}/shutdown'.format(self.SUPERVISOR_API_VERSION),
+                'POST',
+                endpoint=self.RESIN_SUPERVISOR_ADDRESS,
+                api_key=self.RESIN_SUPERVISOR_API_KEY
+            )
+
+    def purge_data(self, app_id, device_id=None):
```
pcarranzav: Might it make sense to leave this just as `purge` so that the methods match the endpoint name? By now IMO we have it well documented that purge refers to /data.

resin/models/supervisor.py
```diff
@@ -0,0 +1,363 @@
...
+        """
+        Clears the user application's /data folder.
+        No need to set device_id and app_id if command is sent to the API on device.
+
+        Args:
+            app_id (str): application id.
+            device_id (Optional[str]): device id.
+
+        Returns:
+            dict: when successful, this dictionary is returned `{ 'Data': 'OK', 'Error': '' }`.
```
pcarranzav, line 238: Does `base_request.request` parse the JSON to return the dictionary?

nghiant2710, line 238: @pcarranzav yup, https://github.com/resin-io/resin-sdk-python/blob/master/resin/base_request.py#L172

resin/models/supervisor.py
nghiant2710, line 105: there is already a wrapper there. IMO, warping this up won't make it look more elegant

resin/models/supervisor.py
nghiant2710: updated!

resin/models/supervisor.py
pcarranzav, line 105: You are repeating the "if self._on_device" check every time, this could be abstracted with a function like:

``` python
def do_request(self, endpoint, data, device_id=None, app_id=None, method='POST'):
    if not self._on_device:
        self._check_args(device_id, app_id)
        data = {
            'deviceId': device_id,
            'appId': app_id,
            'method': method,
            'data': data
        }
            'supervisor/' + endpoint,
            'POST',
            data=data,
            endpoint=self.settings.get('api_endpoint'),
            login=True
        )
    else:
            endpoint,
            method,
            data=data,
            endpoint=self.RESIN_SUPERVISOR_ADDRESS,
            api_key=self.RESIN_SUPERVISOR_API_KEY
        )
```

This would remove around 20 lines of duplicated code on every API-calling method. I will approve if you're convinced this doesn't make it more elegant, but I strongly suggest DRYing it up.

nghiant2710, line 105: @pcarranzav I take my words back. Updated!

resin/models/supervisor.py
```diff
@@ -0,0 +1,290 @@
...
+        if not self._on_device:
+            self._check_args(device_id, app_id)
+
+            data = {
+                'deviceId': device_id,
+                'appId': app_id,
+                'method': 'GET'
+            }
+
+            return self.base_request.request(
```
pcarranzav, line 114: I think you missed the _do_command change on this one?

nghiant2710, line 114: Nope, I find this one a bit odd (different endpoint, passing `method` in request data...) in comparison with others. So I want to leave it alone while wrapping others with _do_command

pcarranzav:
On the supervisor side, LGTM. Took a look at the 2FA part and it looks fine, but you might be better off if someone else checks it.

nghiant2710:
@dfunckt, Can you review this PR?

resin/models/supervisor.py
```diff
@@ -0,0 +1,290 @@
...
+    SUPERVISOR_API_VERSION = 'v1'
+    RESIN_SUPERVISOR_ADDRESS = os.environ.get('RESIN_SUPERVISOR_ADDRESS')
+    RESIN_SUPERVISOR_API_KEY = os.environ.get('RESIN_SUPERVISOR_API_KEY')
+
+    # _on_device = True, if RESIN_SUPERVISOR_ADDRESS and RESIN_SUPERVISOR_API_KEY env vars are avaiable => can communicate with both supervisor API endpoints on device or Resin API endpoints.
+    # _on_device = False, if RESIN_SUPERVISOR_ADDRESS and RESIN_SUPERVISOR_API_KEY env vars are not avaiable => can only communicate with Resin API endpoints.
+    if RESIN_SUPERVISOR_ADDRESS is None or RESIN_SUPERVISOR_API_KEY is None:
+        _on_device = False
+    else:
+        _on_device = True
```
dfunckt: This block can be written in a more succinct form:

```
_on_device = RESIN_SUPERVISOR_ADDRESS is not None and RESIN_SUPERVISOR_API_KEY is not None
```

or even simpler:

```
_on_device = all([RESIN_SUPERVISOR_ADDRESS, RESIN_SUPERVISOR_API_KEY])
```

resin/__init__.py
```diff
@@ -12,6 +12,7 @@
 from .logs import Logs
 from .settings import Settings
 from .models import Models
+from .twofactor_auth import TwoFactor_Auth
```
dfunckt: I'd rename this to `TwoFactoAuth` (my usual stylistic nitpicks :p)

resin/models/supervisor.py
```diff
@@ -0,0 +1,290 @@
...
+
+        Raises:
+            InvalidOption: if endpoint value is not bool.
+
+        """
+
+        try:
+            self._on_device = bool(endpoint)
+        except ValueError:
+            exceptions.InvalidOption('endpoint')
```
dfunckt: `_on_device` is declared as a class attribute, so setting `self._on_device` will change the value for _this particular_ Supervisor instance. Do you happen to create other instances of this class? Or there's only one?

Also, I wouldn't worry about catching the `ValueError` exception -- if someone passes something that can't be coerced to bool, it's an error that should be fixed at programming time.

resin/models/supervisor.py
```diff
@@ -0,0 +1,290 @@
...
+            InvalidOption: if the endpoint is Resin API proxy endpoint and device_id or app_id is not specified.
+
+        """
+
+        self._do_command(
+            '{0}/reboot'.format(self.SUPERVISOR_API_VERSION),
+            device_id=device_id,
+            app_id=app_id,
+            method='POST'
+        )
```
dfunckt, line 202: You say it returns a dict in the method's docstring, but there's no `return something` :p

dfunckt, line 202: Same goes for a couple of other methods below.

resin/twofactor_auth.py
```diff
@@ -0,0 +1,99 @@
...
+        Check if two-factor authentication is enabled.
+
+        Returns:
+            bool: True if enabled. Otherwise False.
+
+        """
+
+        try:
+            self.token.get_property('twoFactorRequired')
+            return True
```
dfunckt, line 33: I assume `get_property` returns a string, right? To be on the safe side, I'd suggest you try to parse the value whether it's '1', 'true' or 'on' and not default to returning `True` just because the property exists -- it can accidentally be set to a falsy value

dfunckt, line 33: I read the JS SDK source and it now makes sense. Ignore my previous comment.

resin/models/supervisor.py
nghiant2710: fixed!

resin/__init__.py
nghiant2710: This name makes sense to me, fixed!

resin/models/supervisor.py
nghiant2710, line 202: fixed!

resin/models/supervisor.py
nghiant2710: @dfunckt  this is the only Supervisor instance so no need to worry about `_on_device`. Removed catching exeception

dfunckt:
LGTM!

PR #21, opened 2015-11-26
Fixes #18: set return value of resin.auth.is_logged_in() to bool

### LINKED ISSUE #18: auth.is_logged_in() return value is wrong
This `auth.is_logged_in()` function is returning auth token not a boolean value as describing.

### LINKED ISSUE #19: Change `RESIN_API_KEY` name since it's reserved and can't be set in the Resin Dashboard
Change `RESIN_API_KEY` name since it's reserved and can't be set in the Resin Dashboard

nghiant2710:


nghiant2710:
@dfunckt, can you review this?

resin/auth.py
```diff
@@ -105,9 +105,13 @@ def is_logged_in(self):
...
 
-        return self.base_request.request(
-            '/whoami', 'GET', endpoint=self.settings.get('api_endpoint')
-        )
+        try:
+            self.base_request.request(
+                '/whoami', 'GET', endpoint=self.settings.get('api_endpoint')
+            )
+            return True
+        except:
```
dfunckt: What exceptions `base_request` throws on error? I'd rather you only caught these specific exceptions, not _every_ exception, as it's now (a plain `except:` catches all exceptions, regardless).

nghiant2710: @dfunckt this method will check if the user is logged or not (haven't logged in, expired token, invalid api key are considered as not logged in), so basically if that API call doesn't raise any error, it means user is logged in. That's why a general catch is fine here.

dfunckt: I understand that :)

However, if the token has expired, which exception is thrown? Is it Python's general `Exception`, an exception from the `requests` module, an exception subclass of yours?

All I'm saying is that you should catch these specific exceptions there, not irrelevant exceptiosn, like one resulting from bad network -- these must be able to propagate.

nghiant2710: @dfunckt agree, I fixed this :)

dfunckt: Thanks! Looks good :+1:

PR #24, opened 2016-01-11
Fix #22 error when reading settings file and #23 bug in device.get_local_ip_address() function.

### LINKED ISSUE #22: Handle errors when reading settings
It's possible that reading `~/.resin/resin.cfg` will fail and the reading code is not robust. It currently checks for the existence of the data directory and .cfg file before attempting to read settings, but it's possible that the file exists but _hasn't_ been saved correctly. Here's a traceback that reads the file but it's not in proper format:

```
    a = resin.Auth()
  File "/var/projects/foo/virtualenv/local/src/resinsdk/resin/auth.py", line 14, in __init__
  File "/var/projects/foo/virtualenv/local/src/resinsdk/resin/base_request.py", line 23, in __init__
    self.settings = Settings()
  File "/var/projects/foo/virtualenv/local/src/resinsdk/resin/settings.py", line 41, in __init__
    self.__read_settings()
  File "/var/projects/foo/virtualenv/local/src/resinsdk/resin/settings.py", line 64, in __read_settings
    options = config_reader.options(self.CONFIG_SECTION)
  File "/usr/lib/python2.7/ConfigParser.py", line 279, in options
    raise NoSectionError(section)
NoSectionError: No section: 'Settings'
```

Unless we save user data in resin.cfg, I think it makes sense to try/catch reading the settings, and fallback to just writing the file, like so:

```
    def __init__(self):
        try:
                self.__read_settings()
        except Exception as e:
                print('Settings file not found or not in proper format', e)
                self.__write_settings()
```

nghiant2710:


resin/settings.py
```diff
@@ -35,15 +36,20 @@ class Settings(object):
                                             'cache')
 
     def __init__(self):
-        if Path.isdir(self._setting['data_directory']):
+        try:
+            self.__read_settings()
+        except:
+            # move old settings file if it exists.
             if Path.isfile(Path.join(self._setting['data_directory'],
                                      self.CONFIG_FILENAME)):
```
dfunckt: The config's path is used several times in this method, so you could assign it once to a local variable and use that. It'll make the code easier to read. `config_path = Path.join(self._setting['data_directory'], self.CONFIG_FILENAME)`

resin/settings.py
```diff
@@ -35,15 +36,20 @@ class Settings(object):
...
-                self.__read_settings()
-            else:
-                self.__write_settings()
-        else:
-            # data directory doesn't exist
+                shutil.move(Path.join(self._setting['data_directory'],
```
dfunckt: There's a race condition here. By the time `shutil.move` executes, the original config file may have been moved and your code will crash. When dealing with the filesystem it's best to try to do something directly (eg. without checking for existence) and handle any errors. I'd rather you wrote it like so:

```
try:
  shutil.move(config_path, '{0}.{1}'.format(config_path, 'old'))
except OSError:
  pass
```

You should then remove the check for the file's existence in the `if` above.

resin/settings.py
```diff
@@ -35,15 +36,20 @@ class Settings(object):
...
+                                     self.CONFIG_FILENAME)
+                            , Path.join(self._setting['data_directory'],
+                                     "{0}.{1}".format(self.CONFIG_FILENAME,'old')))
             self.__write_settings()
+            raise exceptions.InvalidSettings(Path.join(
+                                            self._setting['data_directory'],
+                                            self.CONFIG_FILENAME))
```
dfunckt: Why do you go in all this trouble to recreate the file while keeping a backup and then raise an exception? If it's to warn the user, then warn the user, don't crash :)

resin/settings.py
nghiant2710: @dfunckt I've noticed about this, even a case when we don't have the permissions to write the default settings file again but  there will be an exception thrown in either way (InvalidSettings by us or OSError as you mention) So I don't think we need this catch here

resin/settings.py
nghiant2710: @dfunckt I don't think we should do that while we couldn't correct the settings file. User might have some custom settings and we will go wrong if we write default settings and keep the app running.

dfunckt: I don't see how raising an exception is helpful here. This code will crash once (when it finds sth is wrong with the settings), but will then happily start and keep running if the user runs a command again (because you fixed the problem by writing a proper settings file). You're backing up the previous file, so I think it's more appropriate to issue a warning and keep going.

dfunckt:
I looked through the code and found that you keep creating `Settings()` instances in each model and `BaseRequest()`, so you end up with many settings instances.

This means that settings are not global, which also means that you're overwriting them all over the place and they're never consistent -- they'll be written to file if an instance is changed, but the change won't propagate to other settings instances which will overwrite the file yet again with stale data when saved.

The way you currently use it, either settings must be immutable, or you must make sure you always access the same instance.

resin/settings.py
nghiant2710: Hmm, I only worry about custom settings from user which won't be available if keep the code running but your opinion makes more sense.
I've updated the PR, @dfunckt

nghiant2710:
@dfunckt  I see the issue with current settings, it happens since we cache the settings. Issue created here https://github.com/resin-io/resin-sdk-python/issues/25. Will fix that in the next PR.

nghiant2710:
@dfunckt, is this PR good to go?

dfunckt:
Yes, LGTM. Merge away!

But please have a look at refactoring settings (in another PR).

PR #26, opened 2016-02-02
Implement resin.models.device.get_status()

nghiant2710:


resin/models/device.py
```diff
@@ -17,6 +18,8 @@ class Device(object):
 
     """
 
+    RESIN_CREATION_YEAR = 2013
```
dfunckt: 2013 used below is arbitrary, it doesn't have to be a constant. It could be 2000 and it'll still work. I suggest you inline 2013.

resin/models/device.py
```diff
@@ -687,3 +690,42 @@ def move(self, uuid, app_name):
...
+            str: status of a device. List of available statuses: Idle, Configuring, Updating, Offline and Post Provisioning.
+
+        Examples:
+            >>> resin.models.device.get_status('8deb12a58e3b6d3920db1c2b6303d1ff32f23d5ab99781ce1dde6876e8d143')
+            'Offline'
+
+        """
+
+        device = self.get(uuid)
+        print device
```
dfunckt: leftover

resin/models/device.py
```diff
@@ -687,3 +690,42 @@ def move(self, uuid, app_name):
...
+        if device['provisioning_state'] == 'Post-Provisioning':
+            return DeviceStatus.POST_PROVISIONING
+
+        last_seen = datetime.strptime(device['last_seen_time'], "%Y-%m-%dT%H:%M:%S.%fZ")
```
dfunckt: 'last_seen_time' cannot be None currently due to an API bug, but better guard against this:

```
seen = device['last_seen_time'] and int(datetime.strptime(device['last_seen_time'], '%Y')) < 2013
```

So, the following conditional becomes:

```
if not device['is_online'] and not seen
```

resin/models/device.py
```diff
@@ -687,3 +690,42 @@ def move(self, uuid, app_name):
...
+        if not device['is_online'] and last_seen.year < self.RESIN_CREATION_YEAR:
+            return DeviceStatus.CONFIGURING
+
+        if not device['is_online']:
+            return DeviceStatus.OFFLINE
+
+        if not device['download_progress'] and device['status'] == 'Downloading':
```
dfunckt: Here you have to check for 'download_progress' being None, as it can also be zero and you'll end up with a weird result:

```
if device['download_progress'] is not None and device['status'] == 'Downloading':
```

dfunckt: Note that with what you have above you'll never get the Downloading state, because you test backwards: you should be testing for 'download_progress' being _set_, not `not device['download_progress']`

resin/models/device.py
```diff
@@ -687,3 +690,42 @@ def move(self, uuid, app_name):
...
+            return DeviceStatus.UPDATING
+
+        if not device['download_progress']:
```
dfunckt: You need to be testing for 'provisioning_progress' not being None here.

resin/resources.py
```diff
@@ -25,3 +25,15 @@ class Message(object):
     DEVICE_NOT_WEB_ACCESSIBLE = "Device is not web accessible: {uuid}"
     INCOMPATIBLE_APPLICATION = "Incompatible application: {application}"
     INVALID_SETTINGS = "Settings file not found or not in proper format. Rewriting default settings to: {path}"
+
+
+class DeviceStatus(object):
```
dfunckt: Any reason this class isn't in `device.py`?

resin/models/device.py
nghiant2710: Actually, I don't know why it's 2013. I got it from the js SDK. Updated anyway.

resin/models/device.py
nghiant2710: removed!

resin/models/device.py
nghiant2710: @dfunckt thanks for the pointers. It should become `if not device['is_online'] and seen`

resin/models/device.py
dfunckt: PineJS returns '1970-01-01' for null dates, instead of null -- it's a bug. So, some legacy code used `year < 2013` to filter these dates (since it's impossible to get anything dating before that given that Resin is younger that that!) -- and this code was carried over to other places :)

resin/models/device.py
nghiant2710: Updated!

resin/models/device.py
nghiant2710: Updated!

resin/resources.py
nghiant2710: I see them as constants. However, it makes sense to move them to device.py

resin/models/device.py
```diff
@@ -687,3 +699,41 @@ def move(self, uuid, app_name):
...
+            'Offline'
+
+        """
+
+        device = self.get(uuid)
+        if device['provisioning_state'] == 'Post-Provisioning':
+            return DeviceStatus.POST_PROVISIONING
+
+        seen = device['last_seen_time'] and (datetime.strptime(device['last_seen_time'], "%Y-%m-%dT%H:%M:%S.%fZ")).year < 2013
+        if not device['is_online'] and seen:
```
dfunckt: This must be `if not device['is_online'] and not seen`, or you'll end up inferring that the device is in configuring state instead of simply offline all the time.

nghiant2710:
@dfunckt, I ported this function from resin-sdk without knowing much about the API's response. So I really appreciate your help, updated all your comments :+1:

resin/models/device.py
nghiant2710: @dfunckt let's say a configured device which is offline. So the year of last_seen_time should be greater than 2013 -> seen is `false` -> `if not device['is_online'] and not seen` is true -> status is `CONFIGURING` which is wrong. It's should be `offline`

dfunckt: Yeah, you're right, but for the wrong reasons :p

A device is "seen" if it's last_seen_date is not null. Currently null is represented by '1970-01-01', so the test above must change to read '>=' instead, and the `if` to `not device-is-online and not seen`. Makes sense?

nghiant2710: @dfunckt Updated!

dfunckt:
Awesome, LGTM :+1:

PR #27, opened 2016-02-02
Patch device types to be marked as ALPHA and BETA

nghiant2710:


dfunckt:
The RPi title also needs to be normalised:

```
if dev_type['slug'] == 'raspberry-pi'
    dev_type['name'] = 'Raspberry Pi (v1 or Zero)'
```

And I'd rename the method to `_normalize_device_type`.

dfunckt:
Otherwise, LGTM.

resin/models/config.py
```diff
@@ -41,6 +50,7 @@ def get_all(self):
         if self._config is None:
             self._config = self.base_request.request(
                 'config', 'GET', endpoint=self.settings.get('api_endpoint'))
+            self._config['deviceTypes'] = map(self._patch_device_type, self._config['deviceTypes'])
```
dfunckt: This is a bit fragile -- I think it'll break in Python 3. Better move `_patch_device_type` to a module-level function instead of method, and reference it here instead.

nghiant2710:
@dfunckt, updated! Thanks for the pointers!

dfunckt:
LGTM :+1:

PR #28, opened 2016-02-02
Implement global settings

nghiant2710:


resin/settings.py
```diff
@@ -33,53 +34,61 @@ class Settings(object):
         'token_refresh_interval': (1 * 1000 * 60 * 60)
     }
 
-    _setting['cache_directory'] = Path.join(_setting['data_directory'],
-                                            'cache')
+    _default_setting['cache_directory'] = Path.join(_default_setting['data_directory'],
```
dfunckt, line 37: You can move this inside the hash literal above likse this:

```
'data_directory': Path.join(HOME_DIRECTORY, '.resin'),
'cache_directory': Path.join(HOME_DIRECTORY, '.resin', 'cache'),
...
```

dfunckt:
I think the Settings class is a bit ugly and can be improved. The main issue is the use `@static_method`. You basically have two options, one that is easy to implement but still a bit ugly, the other a bit harder but nice. You can essentially choose between:

1) Moving all Settings class methods to be module-level. You essentially use the Settings class as a module already. That's easy to do, though I don't prefer it over the next option because it robs you flexibility.

2) Study [this class](https://github.com/django/django/blob/master/django/utils/functional.py#L360), extract it (adding a note where you took it from), refactor your Settings class to use `self` properly again (revert essentially to how it previously was) and just create a module-level, lazily initialized object like this: `settings = SimpleLazyObject(lambda: Settings())`

If you need any help, just ping me.

PR #31, opened 2016-04-27
Implement new functions and few improvements for supervisor

nghiant2710:


dfunckt:
Code/Docs-wise LGTM, but let's wait for @pcarranzav to confirm the API calls to device.

DOCUMENTATION.md
```diff
@@ -525,7 +543,7 @@ u'python-sdk-test-device'
 ```
 ### Function: restart(uuid)
 
-Restart a device. This function only works if you log in using credentials or Auth Token.
+Restart an user application container on device. This function only works if you log in using credentials or Auth Token.
```
pcarranzav: Nit: `a` instead of `an`

resin/models/supervisor.py
```diff
@@ -310,7 +333,242 @@ def restart(self, app_id, device_id=None):
...
+            )
+
+    def stop_application(self, app_id=None, device_uuid=None):
+        """
+        Temporarily stops an user application container. Application container will not be removed after invoking this function and a reboot or supervisor restart will cause the container to start again.
+        This function requires supervisor v1.8 or higher.
+        No need to set device_uuid and app_id if command is sent to the API on device.
+
+        Args:
+            app_id (Optional[str]): application id.
```
pcarranzav: `app_id` should be required, not optional, in these and the methods below

DOCUMENTATION.md
nghiant2710: Updated!

resin/models/supervisor.py
nghiant2710: Thanks for pointing this out, @pcarranzav 
Updated!

pcarranzav:
LGTM

PR #32, opened 2016-07-04
Fix issue with PubNub logs channel. If available, use device.logs_cha…

nghiant2710:
…nnel for PubNub channel names

### DIFF
--- a/resin/logs.py
+++ b/resin/logs.py
@@ -3,6 +3,7 @@ from functools import wraps
 
 from .base_request import BaseRequest
 from .models.config import Config
+from .models.device import Device
 
 
 class Logs(object):
@@ -17,6 +18,7 @@ class Logs(object):
 
     def __init__(self):
         self.config = Config()
+        self.device = Device()
 
     def _init_pubnub(func):
         @wraps(func)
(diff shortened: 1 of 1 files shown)


resin/logs.py
```diff
@@ -105,4 +106,12 @@ def get_channel(uuid):
 
         """
 
-        return 'device-{uuid}-logs'.format(uuid=uuid)
+        if not hasattr(self, 'logs_channel'):
+            device_info = self.device.get(uuid)
+            print device_info
```
dfunckt: :)

resin/logs.py
```diff
@@ -105,4 +106,12 @@ def get_channel(uuid):
...
+            if 'logs_channel' in device_info:
+                self.logs_channel = device_info['logs_channel']
+            else:
+                self.logs_channel = uuid
+
+        return 'device-{logs_channel}-logs'.format(logs_channel=self.logs_channel)
```
dfunckt, line 116: This need not be so verbose--it's a single variable. Maybe use the {0} shorthand?

dfunckt:
LGTM

resin/logs.py
nghiant2710: removed :D

resin/logs.py
nghiant2710, line 116: it will give me better context when lookat in the future

Issue #33, opened 2016-08-14
Errors should go to STDERR


imrehg:
Error messages such as [this](https://github.com/resin-io/resin-sdk-python/blob/2130a1270d41cf112d500df23b189f8650882644/resin/settings.py#L55) should go to it's proper place so it can be separated from standard output.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_kie)

PR #34, opened 2016-08-14
print invalid settings error message to STDERR

imrehg:
One possible solution is this, the minimum amount of change to fix things up, related to #33 

An alternative is to completely hide this message, as the problem it reports is actually fixed by the code preceding it, so it just has a little informational value, but no practical effect, if I understand correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## Changed
+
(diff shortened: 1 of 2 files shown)


nghiant2710:
@imrehg sorry, I missed the notification for this PR until today. Thanks a lot :+1: 
LGTM but can you update the changelog before merging please?
IMO, we should keep the message since user can have specific settings and by rewriting default settings, they will be wiped. So it's necessary to have a message.

nghiant2710:
LGTM, will merge it now

PR #37, opened 2016-09-02
changed print statement to fix invalid syntax

ghost:


nghiant2710:
@freddystban we have a fix for it here https://github.com/resin-io/resin-sdk-python/pull/38. This way it can support both Python 2.x and 3.x.
Again, thanks for your PR :+1:

ghost:
Ok, thanks.

PR #47, opened 2017-05-26
Update Python SDK with latest API.

nghiant2710:
This is a breaking PR that updates the Python SDK with latest API development. Core classes are up to date but some functions are skipped  (Marked as TODO) which will be added later.


resin/exceptions.py
```diff
@@ -313,3 +313,63 @@ def __init__(self, required_version, current_version):
...
+    Attributes:
+        code (str): exception code.
+        exit_code (int): program exit code.
+        message (str): error message.
+
+    """
+
+    def __init__(self, application):
+        self.code = 'AmbiguousApplication'
+        self.exit_code = 1
```
dfunckt: What's the point in storing an exit code if it's always "1"? And why is it always "1"?

resin/exceptions.py
```diff
@@ -313,3 +313,63 @@ def __init__(self, required_version, current_version):
         self.code = 'UnsupportedFunction'
         self.exit_code = 1
         self.message = Message.SUPERVISOR_VERSION_ERROR.format(req_version=required_version, cur_version=current_version)
+
+
+class AmbiguousApplication(Exception):
+    """
+    Exception of Resin ambiguous application.
```
dfunckt: This docstring isn't very helpful, is it? :) I mean you merely restate the exception name. Would be better if you described things such as what "ambiguous application" is, when the exception can be raised, etc. Or remove it completely.

resin/exceptions.py
```diff
@@ -313,3 +313,63 @@ def __init__(self, required_version, current_version):
...
+
+    Attributes:
+        code (str): exception code.
+        exit_code (int): program exit code.
+        message (str): error message.
+
+    """
+
+    def __init__(self, application):
+        self.code = 'AmbiguousApplication'
```
dfunckt: What's the reason you're having the exception name stored in an instance var?

resin/models/application.py
```diff
@@ -56,10 +57,13 @@ def get(self, name):
             'eq': name
         }
         try:
-            return self.base_request.request(
+            apps = self.base_request.request(
                 'application', 'GET', params=params,
                 endpoint=self.settings.get('pine_endpoint')
-            )['d'][0]
+            )['d']
+            if len(apps) > 1:
```
dfunckt, line 65: When can this happen?

resin/models/application.py
```diff
@@ -56,10 +57,13 @@ def get(self, name):
...
+                raise exceptions.AmbiguousApplication(name)
```
dfunckt, line 66: Let's document that this method now raises this exception in the docstring.

resin/models/application.py
```diff
@@ -217,27 +221,25 @@ def restart(self, name):
...
         Args:
             name (str): application name.
 
         Returns:
-            str: API key.
+            dict: application config.json content.
 
         Raises:
-            ApplicationNotFound: if application couldn't be found.
+            Request error: if user doesn't have permission to access the application.
```
dfunckt: Not sure if this is controlled by the SDK, but it's usually better to obscure the fact that something is inaccessible by someone by raising a NotFound exception. Although if that's dictated by the API itself, I suggest not changing this.

resin/models/build.py
```diff
@@ -0,0 +1,80 @@
...
+
+        params = {
+            'filter': 'application',
+            'eq': app_id
+        }
+        full_builds = self.base_request.request(
+            'build', 'GET', params=params,
+            endpoint=self.settings.get('pine_endpoint')
+        )
+        if full_builds:
```
dfunckt: When full_builds is false-y? And why return None in that case?

resin/models/device.py
```diff
@@ -60,10 +61,13 @@ def get(self, uuid):
             'eq': uuid
         }
         try:
-            return self.base_request.request(
+            devices = self.base_request.request(
                 'device', 'GET', params=params,
                 endpoint=self.settings.get('pine_endpoint')
-            )['d'][0]
+            )['d']
+            if len(devices) > 1:
```
dfunckt, line 68: When can this happen? AFAIK, device uuids are unique.

resin/models/device.py
```diff
@@ -530,28 +534,6 @@ def register(self, app_name, uuid):
             endpoint=self.settings.get('pine_endpoint'), login=True
         )
 
-    def restart(self, uuid):
```
dfunckt: This is removed completely as far as I can see. Is there a replacement for it?

resin/models/device.py
```diff
@@ -737,3 +719,81 @@ def get_status(self, uuid):
...
+            >>> location = {
+                'latitude': '21.032777',
+                'longtitude': '105.831586'
+            }
+            >>> resin.models.device.set_custom_location('df09262c283b1dc1462d0e82caa7a88e52588b8c5d7475dd22210edec1c50a',location)
+            OK
+
+        """
+
+        if not self.has(uuid):
```
dfunckt, line 766: Apparently this does a request to the API to determine whether a device exists. This is a bit fragile because by the time you make the request below the device might have been deleted. It's better if you tried to patch the device directly and propagate the 404 -- the API will respond appropriately.

resin/models/environment_variables.py
```diff
@@ -27,6 +28,17 @@ def __init__(self):
         self.device = Device()
         self.settings = Settings()
 
+    def __fix_device_env_var_name_key(self, env_var):
```
dfunckt: Double underscore named attributes are reserved by Python. Better use a single underscore to mark it as internal.

resin/models/environment_variables.py
```diff
@@ -27,6 +28,17 @@ def __init__(self):
...
+        """
+        Internal method to workaround the fact that applications environment variables contain a `name` property
+        while device environment variables contain an `env_var_name` property instead.
+        """
+
+        if 'env_var_name' in env_var.keys():
```
dfunckt: Can be simply "if 'env_var_name' in env_var:"

dfunckt (changes requested):

resin/exceptions.py
nghiant2710: Meaningless descriptions removed.

resin/exceptions.py
nghiant2710: Hmm, I think in the past I did follow the way the JS SDK defines any exceptions so we have code, exit code and message for each exception type.

resin/exceptions.py
nghiant2710: My bad since I don't have more than one exit code, exit code is reserved for future use and we will definitely add more exit code and definition for each later.

resin/models/application.py
nghiant2710: updated

resin/models/application.py
nghiant2710: https://github.com/resin-io/resin-sdk/blob/master/lib/models/application.coffee#L146-L147. It's added in JS SDK.
This can happen when you log in as an admin but not sure how the user can have this (I checked our dashboard and we do have a check for application name)

resin/models/application.py
nghiant2710: Updated

resin/models/build.py
nghiant2710: my bad, I checked the API response again and full_builds will never be false so None will never be returned but we don't need `if` statement here. Removed

resin/models/device.py
nghiant2710: The same as application, it's added in JS SDK https://github.com/resin-io/resin-sdk/blob/master/lib/models/device.coffee#L293-L294

resin/models/device.py
nghiant2710: we have `resin.models.supervisor.restart()` as a replacement

resin/models/device.py
nghiant2710: the API always returns `OK` even I patch a device that I don't have access so I will keep this check.

resin/models/environment_variables.py
nghiant2710: My bad, I copied the function name from JS SDK and forgot to remove the extra underscore. Updated.

resin/models/environment_variables.py
nghiant2710: Updated

nghiant2710:
@dfunckt I've updated the PR. Can you please check it again?

resin/exceptions.py
dfunckt: I don't see the value of it, but it's fine with me. BTW, assuming `self.code` is always the exception class name, you can do `self.code = self.__name__`.

resin/exceptions.py
dfunckt: Not sure I understand this bit: "I don't have more than one exit code". What's blocking us from having distinct exit codes for each exception?

resin/models/application.py
dfunckt: Hmm, indeed it's possible when a user is a collaborator to other apps.

How would a user get the specific device they need if getting by name raises `AmbiguousApplication`? Looks like a dead end to me. The JS SDK also accepts an app ID presumably for this reason, you might want to offer the same functionality.

resin/models/device.py
dfunckt: It can't be for the same reason as with apps -- app names are unique per user but device UUIDs are unique globally. I'm assuming the JS SDK has this check for code correctness since the API returns an array. IMO it's redundant, but I'm fine with it in any case.

resin/models/device.py
dfunckt: Why introduce a new entity to the user (i.e. the supervisor)? Also, semantically seems to me as incorrect: you don't want to restart the supervisor, but the application on a specific or all devices.

Is this on parity with the JS SDK (I couldn't find the respective method)?

resin/models/device.py
dfunckt: Hmm, ok 👍 We might want to rethink this a bit on the API side.

dfunckt (changes requested):

resin/exceptions.py
nghiant2710: I meant I've not had time to clearly define other exit codes for every case so for now, only 1 as exit code for all errors

resin/models/application.py
nghiant2710: Yeah, I will add support for both name and id for application and device soon since we now need to prioritize Python 3 SDK support.

resin/exceptions.py
nghiant2710: makes sense, updated 👍

resin/models/device.py
nghiant2710: Yeah, I was wrong, only the function name was changed. I removed the commit for this change. @dfunckt can you check the PR again?

resin/exceptions.py
```diff
@@ -17,7 +17,7 @@ class MissingOption(Exception):
     """
 
     def __init__(self, option):
-        self.code = 'MissingOption'
+        self.code = self.__name__
```
dfunckt: This should be `self.__class__.__name__` apparently.

nghiant2710: @dfunckt updated.

dfunckt (approved):

dfunckt:
BTW I'd suggest you also get approval from one of the JS SDK guys.

resin/exceptions.py
```diff
@@ -310,6 +310,60 @@ class UnsupportedFunction(Exception):
     """
 
     def __init__(self, required_version, current_version):
-        self.code = 'UnsupportedFunction'
+        self.code = self.__class__.__name__
```
pimterry, line 313: Nit: Should this (both `self.code` and `self.exit_code`) live in some kind of `ResinException` superclass?

resin/models/application.py
```diff
@@ -217,27 +222,27 @@ def restart(self, name):
...
         """
 
-        app = self.get(name)
+        # Application not found will be raised if can't get app by id.
+        self.get_by_id(app_id)
         return self.base_request.request(
-            'application/{0}/generate-api-key'.format(app['id']), 'POST',
-            endpoint=self.settings.get('api_endpoint'), login=True
+            '/download-config?appId={id}'.format(id=app_id), 'GET',
```
pimterry: Nit: Why not use the built-in `params` support, rather than building a URL query string by hand?

resin/models/build.py
```diff
@@ -0,0 +1,79 @@
...
+            endpoint=self.settings.get('pine_endpoint')
+        )
+        builds = []
+        for i in full_builds['d']:
+            builds.append({
+                k: i[k] for k in (
+                    'id', 'created_at', 'status',
+                    'push_timestamp', 'end_timestamp',
+                    'start_timestamp', 'project_type',
+                    'commit_hash', 'message', 'log'
```
pimterry: Why do we have a whitelist of fields here, but not in `.get()` above?

resin/models/device.py
```diff
@@ -737,3 +741,81 @@ def get_status(self, uuid):
...
+            raise exceptions.DeviceNotFound(uuid)
+
+        params = {
+            'filter': 'uuid',
+            'eq': uuid
+        }
+        data = {
+            'custom_latitude': location['latitude'],
+            'custom_longitude': location['longtitude']
```
pimterry: There's a breaking typo in this param name here, in the example code above, and `unset` below, but in not the parameter docs. `longtitude` should be `longitude`.

pimterry (changes requested):

resin/exceptions.py
nghiant2710: Makes sense, I will include it in the next PR (after Python3 support).

resin/models/application.py
nghiant2710: this is just the API endpoint,  we have this class https://github.com/resin-io/resin-sdk-python/blob/master/resin/base_request.py to build the URL.

resin/models/build.py
nghiant2710: I've not implement `options` for every functions so can't get in from `.get()`. There fields are the same with this one https://github.com/resin-io/resin-sdk/blob/master/lib/models/build.coffee#L102-L114

nghiant2710:
@pimterry I've updated your comments, can you please have a look again?

resin/models/device.py
nghiant2710: @pimterry Thanks a lot for pointing this out. Updated!

resin/models/build.py
pimterry: Ah, no, you've misunderstood the JS there. What that code is trying to do (I think) is make `.get()` return everything, and `.getAllByApplication` return everything except the `log` property, because the full build logs can be huge. Here you've included `log` (note that it's explicitly commented out in the JS code), which means this code doesn't really do anything, and means that this method will be much slower than it needs to be (since it downloads all the logs for every build ever for an application).

I think you want to change the code in this loop, to instead take the build object, and just remove the the `log` property. That doesn't really solve the problem that the JS code has solved, because you'll still have to download all the logs before you get to this line, but it lets us solve it later. It'd be better to pass proper pine options if you can, but if you can't at least this way users won't start depending on this method returning builds with `.log`, and we can stop transferring them entirely once you've added more pine option support.

resin/models/device.py
pimterry: The JS SDK does it because it's doing `startswith: $uuid`, not `eq`, so you really can easily get multiple results. It is unnecessary here, but I'd leave it anyway for now, since we'll want to make this do `startswith` queries too eventually.

resin/models/application.py
pimterry: I don't think we're talking about the same thing. I mean why are you doing:

```python
'/download-config?appId={id}'.format(id=app_id)
```

instead of

```python
'/download-config', params={ 'app_id': app_id }
```
which is what you do for every other API call here. Minor point, but it's nice to be consistent.

pimterry (changes requested):

resin/models/application.py
nghiant2710: @pimterry, my bad. Since copied the API endpoint from this commit https://github.com/resin-io/resin-sdk/commit/35363e97ae4a4445bbc6f7e7fa8e5716260e26e2 so didn't notice that. Will update it now

resin/models/build.py
nghiant2710: @pimterry, I added `log` on purpose since I think it's mandatory (IMO, it's the main reason they check previous builds on the dashboard). But you make sense since it's huge and annoying sometimes, however since I've not implemented the pine options and we still have to download all the logs, I will add an option to include the log in the return or not. So users will be able to choose what they want to see.

nghiant2710: @pimterry updated!

resin/models/application.py
nghiant2710: updated!

resin/models/build.py
```diff
@@ -0,0 +1,81 @@
...
+        build = self.base_request.request(
+            'build', 'GET', params=params,
+            endpoint=self.settings.get('pine_endpoint')
+        )
+        if build:
+            return build
+        else:
+            raise exceptions.BuildNotFound(id)
+
+    def get_all_by_application(self, app_id, build_log=None):
```
pimterry: Nit: I'd call this something that sounds like a boolean (`include_logs`), and make it properly `False`.

resin/models/build.py
```diff
@@ -0,0 +1,81 @@
...
+        )
+        builds = []
+        build_info = [
+            'id', 'created_at', 'status',
+            'push_timestamp', 'end_timestamp',
+            'start_timestamp', 'project_type',
+            'commit_hash', 'message'
+        ]
+        if build_log:
+            build_info.append('log')
```
pimterry, line 78: The JS code only lists all the fields because it has to for the pine query. You don't have to, so you could avoid that and avoid duplicating the list of fields in the model. It doesn't make much difference if you're adding pine options really soon, but if not, or if you want this to be clearer in the meantime, you could replace everything after `full_builds` with:

```python
builds = full_builds['d']
if include_logs:
  return builds
else:
  return { k: v for build in builds for k, v in build.items() if k is not 'log'}
```

pimterry (approved):
Some last notes, but nothing that's actually _necessary_, LGTM

PR #53, opened 2017-11-23
Update to the latest API version and implement new API endpoints

nghiant2710:
This includes many breaking changes including the upgrade to the API v3.

I really want to break this PR into  smaller ones but since the implementation of new API endpoints all depends on the commit that updates the API to latest version so it's better to keep it as one.


resin/models/application.py
```diff
@@ -311,3 +313,152 @@ def disable_rolling_updates(self, app_id):
...
+            endpoint=self.settings.get('pine_endpoint')
+        )
+
+    def grant_support_access(self, app_id, valid_period):
+        """
+        Grant support access to an application until a specified time.
+
+        Args:
+            app_id (str): application id.
+            valid_period (int): valid period in hour.
```
pimterry: `hour` should probably be `hours`

This is a different approach to the JS SDK though, which takes a timestamp instead. It's not a big deal, but that approach does feel a little more flexible, and it'd be nice to be consistent. Any reason not to make these match up?

resin/models/device.py
```diff
@@ -878,3 +871,92 @@ def generate_device_key(self, uuid):
...
+            '/devices/{}/summary'.format(uuid)
+        )
+
+    def grant_support_access(self, uuid, valid_period):
+        """
+        Grant support access to a device until a specified time.
+
+        Args:
+            uuid (str): device uuid.
+            valid_period (int): valid period in hour.
```
pimterry: Same as above

resin/models/device_os.py
```diff
@@ -24,6 +24,26 @@ def __init__(self):
...
+
+        """
+
+        data = {
+            'appId': app_id
+        }
+        return self.base_request.request(
+            'download-config', 'POST', data=data,
+            endpoint=self.settings.get('api_endpoint')
+        )
```
pimterry, line 45: I think you almost certainly want to allow more options here too. The JS SDK adds the appId, but also whatever else you want to include: https://github.com/resin-io/resin-sdk/blob/45c0ed0cb1af4833952d84b03cbd8a8ed2d30e28/lib/models/os.coffee#L369. As is, you can't use this to configure wifi/network settings/poll intervals etc.

pimterry (changes requested):
A couple of things to look at, but nothing major.

resin/models/application.py
nghiant2710: @pimterry, I select hours as an unit since we use it in our dashboard and IMO, timestamp is a bit harder for end-users to use. Any reason why we don't use the unit that we use on the dashboard? Anyway, I'm fine with timestamp so the final call is up to you.

resin/models/device_os.py
nghiant2710: Yeah, I've seen we support options in the JS SDK and I will implement it (not just for this method) in the next PR. This PR will focus on updating the SDK with latest API.

resin/models/application.py
lurch: Would setting the parameter as a timestamp (instead of as a number of hours) mean that the user has to deal with timezone conversions?

resin/models/application.py
```diff
@@ -311,3 +313,152 @@ def disable_rolling_updates(self, app_id):
...
+        Examples:
+            >> > resin.models.application.grant_support_access('5685', 2)
+            'OK'
+
+        """
+
+        if not valid_period or int(valid_period) <= 0:
+            raise exceptions.InvalidParameter('valid_period', valid_period)
+
+        expiry_timestamp = ((datetime.utcnow() - datetime.utcfromtimestamp(0)).total_seconds() + int(valid_period) * 3600) * 1000
```
lurch: I just ran this code in a REPL, and it returns a float. Is the API okay with that, or should `expiry_timestamp` be converted to an `int`?

resin/models/application.py
pimterry: > Would setting the parameter as a timestamp (instead of as a number of hours) mean that the user has to deal with timezone conversions?

Should be fine. In JS at least it's pretty easy to do timezone agnostically: you can use either `Date.now() + X` or `Date.parse('2017-11-29T14:40:00')`, and you'll get sensible results.

> I select hours as an unit since we use it in our dashboard and IMO, timestamp is a bit harder for end-users to use. Any reason why we don't use the unit that we use on the dashboard? Anyway, I'm fine with timestamp so the final call is up to you.

Imo it's simpler this way as it's a closer representation of what the API is really doing, while still being easy to map into any UI representation you like (including relative times). This means that the API requests are closer to the SDK methods, and you can do things like set the timestamp, and then read it from `is_accessible_by_support_until__date`, and see the same value.

It's just a bit simpler and more predictable than relative time, and it's easy to convert timestamps to relative time if you want to. If there's no strong reasons not to, I'd rather stick with timestamps.

resin/models/device_os.py
pimterry: Alright, fair enough. I don't think this method is super useful without that, but adding more options can wait, it's not a problem.

resin/models/application.py
nghiant2710: @pimterry, makes sense. I've changed to timestamp. Can you check again?

resin/models/application.py
nghiant2710: @lurch yeah, the API is okay but it should be converted. However, I've changed the valid period to timestamp so this no longer is a prolem

pimterry (approved):
:+1:

resin/models/device.py
```diff
@@ -878,3 +871,90 @@ def generate_device_key(self, uuid):
...
+        Revoke support access to a device.
+
+        Args:
+            uuid (str): device uuid.
+
+        Returns:
+            OK.
+
+        Examples:
+            >> > resin.models.device.revoke_support_access('5685')
```
lurch: I guess this example ought to be using a uuid rather than an app_id ?

nghiant2710: @lurch ah yes, my bad. Thanks for pointing this out, updated!

resin/models/application.py
```diff
@@ -311,3 +313,150 @@ def disable_rolling_updates(self, app_id):
...
+        data = {
+            'is_accessible_by_support_until__date': expiry_timestamp
+        }
+
+        return self.base_request.request(
+            'application', 'PATCH', params=params, data=data,
+            endpoint=self.settings.get('pine_endpoint')
+        )
+
+    def revoke_support_access(self, app_id):
```
lurch, line 409: I know nothing about the resin API, but just out of curiosity are there functions you can use to check if support access is enabled for a particular application or device?

nghiant2710: @lurch, we don't have an endpoint for it but we can check `support_expiry_date` property in every device to see whether it allows support access or not. We don't provide this function in both SDK atm but it's definitely a good point. Created an issue for it https://github.com/resin-io/resin-sdk-python/issues/54, I will implement it later.

Issue #45, opened 2017-02-08
Add support for Python3


nghiant2710:
<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_1dph)

imrehg:
@nghiant2710 any update for this?

goodes:
I just hacked a version that uses six and also properly decodes the request response. As there are no unit tests I am unsure how good my fix is. Does anyone want to test this?

It is in the branch 'add-py3-six-support'

https://github.com/goodes/resin-sdk-python/tree/add-py3-six-support

lurch:
Fixed in #55

PR #55, opened 2017-12-03
Add Python 3 support.

### LINKED ISSUE #35: On Python3 the SDK also needs urlparse
While on Python 2 it works fine, the SDK seems to have a missing dependency on Python 3.

```
Python 3.5.2 (default, Jun 28 2016, 08:46:01) 
[GCC 6.1.1 20160602] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import resin
  File "<stdin>", line 1, in <module>
  File "/home/greg/resin/prog/bluemix-resin-python/venv3/lib/python3.5/site-packages/resin/__init__.py", line 9, in <module>
  File "/home/greg/resin/prog/bluemix-resin-python/venv3/lib/python3.5/site-packages/resin/base_request.py", line 7, in <module>
    from urlparse import urljoin
ImportError: No module named 'urlparse'
```

Front logo [Front conversations](https://app.frontapp.com/open/top_12i3)

nghiant2710:
This PR contains some changes to make sure the right libraries and
dependencies are loaded to support Python 3. There is no big change in the source code since most
of the code is Python 3 compatible. Fixes #35 and #45.

Apart from that there are some improments and bug fixes included:
- Implement base class for resin exceptions.
- Fix a bug in `resin.models.device.get_all_by_application()` to make it
works with API v3.

### DIFF
--- a/resin/base_request.py
+++ b/resin/base_request.py
@@ -1,10 +1,12 @@
-import requests
+try:  # Python 3 imports
+    from urllib.parse import urljoin
+except ImportError:  # Python 2 imports
+    from urlparse import urljoin
+
 import json
-import urllib
 from string import Template
 import os
-
-from urlparse import urljoin
+import requests
 
 from .settings import Settings
 from .token import Token
@@ -41,9 +43,8 @@ class BaseRequest(object):
         self.__set_content_type(headers, 'application/json')
         if not stream:
             return requests.post(url, data=json.dumps(data), headers=headers)
-        else:
-            return requests.post(
-                url, data=json.dumps(data), headers=headers, stream=stream)
+        return requests.post(
+            url, data=json.dumps(data), headers=headers, stream=stream)
 
     def __put(self, url, headers, data=None, stream=None):
         self.__set_content_type(headers, 'application/json')
@@ -105,7 +106,7 @@ class BaseRequest(object):
 
         headers = headers or {}
 
-        METHODS = {
+        methods = {
             'get': self.__get,
             'post': self.__post,
             'put': self.__put,
@@ -113,7 +114,7 @@ class BaseRequest(object):
             'head': self.__head,
             'patch': self.__patch
         }
-        request_method = METHODS[method.lower()]
+        request_method = methods[method.lower()]
         url = urljoin(endpoint, url)
         if auth:
             if not api_key:
@@ -170,11 +171,11 @@ class BaseRequest(object):
             raise exceptions.RequestError(response._content)
 
         try:
-            json = response.json()
+            json_data = response.json()
         except ValueError:
             return response.content
 
-        return json
+        return json_data
 
     def _request_new_token(self):
         headers = {}
--- a/resin/exceptions.py
+++ b/resin/exceptions.py
@@ -2,7 +2,22 @@
 from .resources import Message
 
 
-class MissingOption(Exception):
+class ResinException(Exception):
+    """
+    Exception base class for Python SDK.
+
+    Attributes:
+        code (str): exception code.
+        exit_code (int): program exit code.
+
+    """
+
+    def __init__(self):
+        self.code = self.__class__.__name__
+        self.exit_code = 1
+
+
+class MissingOption(ResinException):
     """
     Exception type for missing option in settings or auth token.
 
@@ -10,19 +25,16 @@ class MissingOption(Exception):
         option (str): option name.
 
     Attributes:
-        code (str): exception code.
-        exit_code (int): program exit code.
         message (str): error message.
 
     """
 
     def __init__(self, option):
-        self.code = self.__class__.__name__
-        self.exit_code = 1
+        super(MissingOption, self).__init__()
         self.message = Message.MISSING_OPTION.format(option=option)
 
 
-class InvalidOption(Exception):
+class InvalidOption(ResinException):
     """
     Exception type for invalid option in settings or auth token.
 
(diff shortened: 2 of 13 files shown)


resin/settings.py
```diff
@@ -9,6 +8,11 @@
 from . import exceptions
 from .resources import Message
 
+try:
+    import configparser
```
lurch: Could you add the `# Python 3 imports` / `# Python 2 imports` comments as you've done in the other files?

resin/models/device.py
```diff
@@ -141,7 +144,7 @@ def get_all_by_application_id(self, appid):
 
         """
         params = {
-            'filter': 'application',
+            'filter': 'belongs_to__application',
```
lurch, line 147: Looks like this is a bugfix unrelated to adding Python3 support?

resin/exceptions.py
```diff
@@ -2,222 +2,204 @@
 from .resources import Message
 
 
-class MissingOption(Exception):
+class ResinException(Exception):
```
lurch, line 5: Adding a `ResinException` baseclass obviously isn't strictly related to adding Python3, but it's nice to see anyway. This change should probably be mentioned in the commit-message?

resin/base_request.py
```diff
@@ -143,8 +144,8 @@ def request(self, url, method, endpoint, params=None, data=None,
                     raise exceptions.Unauthorized()
 
             if self.util.should_update_token(
-                self.token.get_age(),
-                self.settings.get('token_refresh_interval')
+                    self.token.get_age(),
+                    self.settings.get('token_refresh_interval')
```
lurch: Why the extra indent here?

resin/settings.py
```diff
@@ -74,18 +78,18 @@ def __write_settings(self, default=None):
 
         if default:
             self._setting = Settings._setting
-        config = ConfigParser.ConfigParser()
+        config = configparser.ConfigParser()
         config.add_section(self.CONFIG_SECTION)
         for key in self._setting:
-            config.set(self.CONFIG_SECTION, key, self._setting[key])
+            config.set(self.CONFIG_SECTION, key, str(self._setting[key]))
```
lurch: There's no code inside `__read_settings` which turns any values into a non-string value, so it seems a bit weird that `__write_settings` is specifically coercing all values to strings. For consistency, perhaps it'd make sense to initialise all the values in `self._setting` to strings upon creation, rather than deferring this step until `__write_settings` ?
i.e. change line 42 to `'image_cache_time': str(1 * 1000 * 60 * 60 * 24 * 7),` etc. and get rid of the `str` here.

setup.py
```diff
@@ -58,6 +58,8 @@ def get_version(package):
         'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.6',
         'Programming Language :: Python :: 2.7',
+        'Programming Language :: Python :: 3.5',
```
lurch: Any reason to not also support Python 3.4 ? (Or 3.3?)
Should there also be a `'Programming Language :: Python :: 3',` ?

resin/base_request.py
nghiant2710: Removed!

resin/exceptions.py
nghiant2710: my bad, I've updated the commit message to clarify this.

resin/models/device.py
nghiant2710: my bad, I've updated the commit message to clarify this.

resin/settings.py
nghiant2710: Updated!

setup.py
nghiant2710: I've not done the test on Python 3.4 or older so I'm not sure if anything can go wrong. I think we can stay with Python v3.5 and v3.6 for now. Once I finish the test suite for the SDK, we can add support for other versions.

lurch:
Maybe you should also add python3.5 and python3.6 to the `.travis.yml` config? :)

resin/settings.py
nghiant2710: I did this because user can add a setting and it can be a non-string value but after checking the code again, I already did convert everything to string before adding so you're right, we should change this.

lurch:
...and perhaps the PR description should be updated to say that this fixes #35 and #45 ?

nghiant2710:
@lurch, I've fixed all your comments, can you check the PR again please?

tests/pep8-git-pr-checker
```diff
@@ -20,17 +20,17 @@ def main():
     files = []
     output = system('git', 'ls-files')
```
lurch: Probably better to add the `.decode("utf-8")` here, and then the below changes (with the exception of `file` -> `open` change) can probably all be reverted?  (because `tmp_file` and `files` will then be `str` rather than `bytes`)

nghiant2710: @lurch Makes sense, I've reverted them all expect `file -> open` since `file` is not available in Python 3 and `open` is the canonical way now.

lurch:
This all LGTM now, but I won't "approve" this PR as I've not run any of the code in this repo and I've never even used the Resin API. I'll request Tim, since he approved your other SDK changes :)

resin/settings.py
```diff
@@ -35,9 +39,9 @@ class Settings(object):
         'api_version': 'v3',
         'data_directory': Path.join(HOME_DIRECTORY, '.resin'),
         # cache time : 1 week in milliseconds
-        'image_cache_time': (1 * 1000 * 60 * 60 * 24 * 7),
+        'image_cache_time': str(1 * 1000 * 60 * 60 * 24 * 7),
         # token refresh interval: 1 hour in milliseconds
-        'token_refresh_interval': (1 * 1000 * 60 * 60)
+        'token_refresh_interval': str(1 * 1000 * 60 * 60)
```
pimterry, line 46: Why are these strings? It looks like everywhere they're used, we really want integers.

I'm guessing it's because when we read them from a config file, they'd end up as strings anyway, and consistency is good (totally agree), but if that's the case, we should briefly comment this above ('all config values must be strings because...), or we're definitely going to forget that and add non-string config items in here in future.

pimterry (changes requested):
One brief point, but everything else looks good to me.

resin/settings.py
nghiant2710: @pimterry makes sense, I've added a comment about this.

pimterry (approved):

Issue #59, opened 2018-01-05
`set_to_build` command does not work


LucianBuzzo:
See https://app.frontapp.com/open/cnv_gtna5t
```
File "resin_cli.py", line 130, in main
res = resin.models.device.set_to_build(arguments["--device_uuid"],build_id)
File "/usr/local/lib/python3.6/site-packages/resin/models/device.py", line 721, in set_to_build
endpoint=self.settings.get('pine_endpoint')
File "/usr/local/lib/python3.6/site-packages/resin/base_request.py", line 171, in request
raise exceptions.RequestError(response._content)
resin.exceptions.RequestError: b'Internal Server Error'
```

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_67nt)

Issue #57, opened 2018-01-02
New API calls


FRVisser:
Hi @nghiant2710,

The new API implementation works confusing because of the login state. If already logged in on the SDK version 1.x the old API address is retained even when updated to the latest 2.0 version. Only after a logout or removal of the resin.cfg file. But, if I try to reproduce it, I do get:
```
Settings file not found or not in proper format. Rewriting default settings to: /Users/username/.resin/resin.cfg
```

2nd point. The `device.set_to_build` command doesn't work. It returns: 
```
>>> resin.models.device.set_to_build("ef448bdb8ac0000bb9013d98c6491587","339379")
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    resin.models.device.set_to_build("ef448bdb8ac0000bb9013d98c6491587","339379")
  File "/usr/local/lib/python2.7/site-packages/resin/models/device.py", line 721, in set_to_build
    endpoint=self.settings.get('pine_endpoint')
  File "/usr/local/lib/python2.7/site-packages/resin/base_request.py", line 171, in request
    raise exceptions.RequestError(response._content)
RequestError
```

3th and last point 😅. Some return values have different labelling with API v3. 
Like:

`build` ➡️  `should_be_running__build`
`commit` ➡️ `is_on__commit`

I understand I am quite early using these commands. However, it would be nice to have some documentation on which values changed in API v3.

Thanks! 😬

nghiant2710:
@FRVisser I'm so sorry for the late response. Regarding your questions, #2 will be fixed by PR #60 and it will be merged soon. We also update the SDK docs in #61 which shows the differences of returned object's properties between old and new API.
About the first point, the config file format of v1.7.0 and v2.0.0 will be the same so if you upgrade from v1.7.0 to v2.0.0 then your config file will be retained, it's a normal behavior. In case your config file is missing or in incorrect format (mandatory keys are not there), the SDK will create a backup file of your current config if possible and write new config file with default values.

FRVisser:
Cool thanks!

PR #60, opened 2018-01-12
Update device.set_to_build() to work with latest API.

nghiant2710:
Change the argument from build id to build commit hash since build id is not available on the dashboard.

### DIFF
--- a/resin/models/build.py
+++ b/resin/models/build.py
@@ -43,9 +43,28 @@ class Build(object):
         else:
             raise exceptions.BuildNotFound(id)
 
+    def get_by_commit(self, commit_hash):
+        """
+        Get a specific build by commit hash.
+        """
+
+        params = {
+            'filter': 'commit_hash',
+            'eq': commit_hash
+        }
+
+        build = self.base_request.request(
+            'build', 'GET', params=params,
+            endpoint=self.settings.get('pine_endpoint')
+        )
+        if build and (len(build['d']) > 0):
+            return build['d']
+        else:
+            raise exceptions.BuildNotFound(commit_hash)
+
     def get_all_by_application(self, app_id, include_logs=False):
         """
-        Get a specific build.
+        Get list of builds belong to an application.
 
         Args:
             app_id (str): application id.
--- a/resin/models/device.py
+++ b/resin/models/device.py
@@ -14,6 +14,7 @@ from ..settings import Settings
 from ..token import Token
 from .. import exceptions
 from .application import Application
+from .build import Build
 
 
 # TODO: support both device uuid and device id
@@ -41,6 +42,7 @@ class Device(object):
         self.settings = Settings()
         self.token = Token()
         self.application = Application()
+        self.build = Build()
 
     def get(self, uuid):
         """
@@ -686,21 +688,22 @@ class Device(object):
             endpoint=self.settings.get('pine_endpoint')
         )
 
-    def set_to_build(self, uuid, build):
+    def set_to_build(self, uuid, build_commit_hash):
         """
-        Set a device to specific build id.
+        Set a device to specific build commit hash.
 
         Args:
             uuid (str): device uuid.
-            build (str): build id.
+            build_commit_hash (str): build commit hash.
 
         Raises:
             DeviceNotFound: if device couldn't be found.
             ApplicationNotFound: if application couldn't be found.
+            FailedBuild: if build commit hash points to an error build.
             IncompatibleApplication: if moving a device to an application with different device-type.
 
         Examples:
-            >> > resin.models.device.set_to_build('8deb12a58e3b6d3920db1c2b6303d1ff32f23d5ab99781ce1dde6876e8d143', '123098')
+            >> > resin.models.device.set_to_build('4457d5db93be458270666ef8b8157c66', '950eac2d3efce555490c96e7c9b55c37b385acb6')
             'OK'
 
         """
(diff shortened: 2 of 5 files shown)


lurch:
Out of curiosity, where's the similar function in the JS SDK? https://github.com/resin-io/resin-sdk/blob/master/DOCUMENTATION.md
Neither https://github.com/resin-io/resin-sdk/search?q=should_be_running__build or https://github.com/resin-io/resin-sdk/search?q=should_be_running show up any search results :-/  (although it's more than likely that I'm missing something obvious)

Should the Python SDK and the JS SDK be offering **roughly** the same functionality? @pimterry 

And just for reference, this PR fixes #59 and fixes part of #57

nghiant2710:
@lurch this function was contributed by an user https://github.com/resin-io/resin-sdk-python/commit/9a7c03a9f0709af17990b0f3ed4c51f2b55bfc2c. It seems like we should add it to JS SDK as well since many users are interested in this function.

resin/models/device.py
```diff
@@ -686,34 +688,42 @@ def disable_device_url(self, uuid):
...
             'OK'
 
         """
 
         if not self.has(uuid):
             raise exceptions.DeviceNotFound(uuid)
 
+        app_id = self.get(uuid)['belongs_to__application']['__id']
+        build_id = None
+        for build in self.build.get_all_by_application(app_id):
```
dfunckt: This can bring in an unbounded number of builds, which can be a few. You can fetch a specific build by its commit hash directly with an OData request.

nghiant2710: I see, updated. Thanks @dfunckt

resin/models/device.py
```diff
@@ -686,34 +688,37 @@ def disable_device_url(self, uuid):
...
-            >> > resin.models.device.set_to_build('8deb12a58e3b6d3920db1c2b6303d1ff32f23d5ab99781ce1dde6876e8d143', '123098')
+            >> > resin.models.device.set_to_build('4457d5db93be458270666ef8b8157c66', '950eac2d3efce555490c96e7c9b55c37b385acb6')
             'OK'
 
         """
 
         if not self.has(uuid):
             raise exceptions.DeviceNotFound(uuid)
 
+        build_id = self.build.get_by_commit(build_commit_hash)[0]['id']
```
dfunckt: This can throw an IndexError if there's no build with this commit hash. Maybe catch that and rethrow a BuildNotFound?

dfunckt (approved):
Looks good but please consider my comment about the exception

resin/models/build.py
```diff
@@ -43,9 +43,28 @@ def get(self, id):
...
+        params = {
+            'filter': 'commit_hash',
+            'eq': commit_hash
+        }
+
+        build = self.base_request.request(
+            'build', 'GET', params=params,
+            endpoint=self.settings.get('pine_endpoint')
+        )
+        if build and (len(build['d']) > 0):
```
nghiant2710, line 60: @dfunckt, I've updated the conditions here to make sure it won't return an empty list.

resin/models/device.py
```diff
@@ -686,34 +688,37 @@ def disable_device_url(self, uuid):
             endpoint=self.settings.get('pine_endpoint')
         )
 
-    def set_to_build(self, uuid, build):
+    def set_to_build(self, uuid, build_commit_hash):
```
pimterry, line 691: Note that this is a breaking change - presumably this means we'll bump to 3.0.0?

resin/models/build.py
```diff
@@ -43,9 +43,28 @@ def get(self, id):
...
+            return build['d']
+        else:
+            raise exceptions.BuildNotFound(commit_hash)
+
     def get_all_by_application(self, app_id, include_logs=False):
         """
-        Get a specific build.
+        Get list of builds belong to an application.
```
pimterry, line 67: You've changed the doc string here, but you haven't regenerated the docs.

resin/models/device.py
pimterry: The API doesn't guarantee only one build per commit hash (as far as I can tell), it only guarantees one _successful_ build per commit hash: https://github.com/resin-io/resin-api/blob/master/src/models/resin.sbvr#L691.

I suspect if you have a failed and then a successful build for the same commit hash, this will get the build id for the failed one, which won't work. You need to filter for success here.

pimterry (changes requested):

resin/models/device.py
pimterry: It looks like from #57 maybe actually the old version never worked? If it definitely never worked (so we can guarantee that nobody's relying on it) then I think changing this as a patch release is fine, but if it's possible the previous method worked then we do need to mark this as major.

lurch: ...and it's only a minor nit, but I was wondering if `BuildNotFound(commit_hash)` should actually be a different exception from `BuildNotFound(id)`? Or perhaps that's such a subtle difference that it's not worth worrying about? \*shrug\*

pimterry: I know the JS SDK uses the same exception class for both, although I'd be open to changing it if there's value to doing so... The main reason for separate error classes as I understand it are:

* Catch errors for missing builds specifically (in which case I don't think you care which format was used)
* Show easily understandable errors in the console if they're not caught (ditto)

Is there a use case where you need to know which format you passed from the error itself?

lurch: > I know the JS SDK uses the same exception class for both

Ah, in that case I'm happy for the Python SDK to do the same then :-)

resin/models/build.py
nghiant2710: updated!

resin/models/device.py
nghiant2710: @pimterry, this function works on old pine endpoint (`ewa`) and I'm sure that we have users are using it so yeah, we will bump to v3.0.0 (will merge this PR to v3 branch and release v3 after I finish implementing the api key changes)

resin/models/device.py
nghiant2710: @pimterry makes sense, I've added a check to make sure users can only set to a valid build.

pimterry (approved):

PR #58, opened 2018-01-04
Add VersionBot changes to repository.

hedss:
Connects-to: #56
Change-type: patch

PR #61, opened 2018-01-12
Update Python SDK Docs

nghiant2710:
- Adding `Build` class to the docs.
- Adding details about the old and new returned objects from old and new API.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -10,12 +10,13 @@ hesitate to open an issue in GitHub, we'll be happy to help.
 - [Resin](#resin)
     - [Models](#models)
         - [Application](#application)
+        - [Build](#build)
         - [Config](#config)
         - [Device](#device)
         - [DeviceOs](#deviceos)
         - [EnvironmentVariable](#environmentvariable)
-            - [DeviceEnvVariable](#deviceenvvariable)
             - [ApplicationEnvVariable](#applicationenvvariable)
+            - [DeviceEnvVariable](#deviceenvvariable)
         - [Key](#key)
         - [Supervisor](#supervisor)
     - [Auth](#auth)
@@ -29,6 +30,12 @@ This module implements all models for Resin Python SDK.
 ## Application
 
 This class implements application model for Resin Python SDK.
+
+Due to API changes, the returned Application object schema has changed. Here are the formats of the old and new returned objects.
+
+The old returned object's properties: `__metadata, actor, app_name, application, commit, device_type, git_repository, id, should_track_latest_release, support_expiry_date, user, version`.
+
+The new returned object's properties (since Python SDK v2.0.0): `__metadata, actor, app_name, commit, depends_on__application, device_type, git_repository, id, is_accessible_by_support_until__date, should_track_latest_release, user, version`.
 ### Function: create(name, device_type)
 
 Create an application. This function only works if you log in using credentials or Auth Token.
@@ -284,9 +291,70 @@ Revoke support access to an application.
 #### Examples:
     >> > resin.models.application.revoke_support_access('5685')
     'OK'
+## Build
+
+This class implements build model for Resin Python SDK.
+### Function: get(id)
+
+Get a specific build.
+
+#### Args:
+    id (str): build id.
+
+#### Returns:
+    dict: build info.
+
+#### Raises:
+    BuildNotFound: if build couldn't be found.
+### Function: get_all_by_application(app_id, include_logs)
+
+Get list of builds belonging to an application.
+
+#### Args:
+    app_id (str): application id.
+    include_logs (Optional[bool]): Defaults to False since these may be very large. True if user wants to include build logs in build info.
+
+#### Returns:
+    list: list of build info.
+## Config
+
+This class implements configuration model for Resin Python SDK.
+
+#### Attributes:
+    _config (dict): caching configuration.
+### Function: get_all()
+
+Get all configuration.
+
+#### Returns:
+    dict: configuration information.
+
+#### Examples:
+```python
+>>> resin.models.config.get_all()
+{ all configuration details }
+```
+### Function: get_device_types()
+
+Get device types configuration.
+
+#### Returns:
+    list: device types information.
+
+#### Examples:
+```python
+>>> resin.models.config.get_device_types()
+[ all configuration details ]
+```
 ## Device
 
 This class implements device model for Resin Python SDK.
+
+Due to API changes, the returned Device object schema has changed. Here are the formats of the old and new returned objects.
+
+The old returned object's properties: `__metadata, actor, application, build, commit, created_at, custom_latitude, custom_longitude, device, device_type, download_progress, id, ip_address, is_connected_to_vpn, is_online, is_web_accessible, last_connectivity_event, last_vpn_event, latitude, local_id, location, lock_expiry_date, logs_channel, longitude, name, note, os_variant, os_version, provisioning_progress, provisioning_state, public_address, service_instance, status, supervisor_release, supervisor_version, support_expiry_date, user, uuid, vpn_address`.
+
+The new returned object's properties (since Python SDK v2.0.0): `__metadata, actor, belongs_to__application, belongs_to__user, created_at, custom_latitude, custom_longitude, device_type, download_progress, id, ip_address, is_accessible_by_support_until__date, is_connected_to_vpn, is_locked_until__date, is_managed_by__device, is_managed_by__service_instance, is_on__commit, is_online, is_web_accessible, last_connectivity_event, last_vpn_event, latitude, local_id, location, logs_channel, longitude, name, note, os_variant, os_version, provisioning_progress, provisioning_state, public_address, should_be_managed_by__supervisor_release, should_be_running__build, status, supervisor_version, uuid, vpn_address`.
 ### Function: disable_device_url(uuid)
 
 Disable device url for a device.
@@ -791,36 +859,6 @@ clear custom location for a device.
 ```python
 >>> resin.models.device.unset_custom_location('df09262c283b1dc1462d0e82caa7a88e52588b8c5d7475dd22210edec1c50a')
 OK
-```
-## Config
-
-This class implements configuration model for Resin Python SDK.
-
-#### Attributes:
-    _config (dict): caching configuration.
-### Function: get_all()
-
-Get all configuration.
-
-#### Returns:
-    dict: configuration information.
-
-#### Examples:
-```python
->>> resin.models.config.get_all()
-{ all configuration details }
-```
-### Function: get_device_types()
-
-Get device types configuration.
-
-#### Returns:
-    list: device types information.
-
-#### Examples:
-```python
->>> resin.models.config.get_device_types()
-[ all configuration details ]
 ```
 ## DeviceOs
 
--- a/docs_generator.py
+++ b/docs_generator.py
@@ -13,12 +13,13 @@ Table_Of_Content = [
     ('Resin', TOC_ROOT),
     ('Models', TOC_L1),
     ('Application', TOC_L2),
+    ('Build', TOC_L2),
     ('Config', TOC_L2),
     ('Device', TOC_L2),
     ('DeviceOs', TOC_L2),
     ('EnvironmentVariable', TOC_L2),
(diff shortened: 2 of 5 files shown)


DOCUMENTATION.md
```diff
@@ -284,9 +291,73 @@ Revoke support access to an application.
 #### Examples:
     >> > resin.models.application.revoke_support_access('5685')
     'OK'
+## Build
+
+This class implements model for build for Resin Python SDK.
```
lurch: nit: Could change this to `This class implements build model for Resin Python SDK.` to match the earlier `This class implements application model for Resin Python SDK.` phrasing?

DOCUMENTATION.md
```diff
@@ -29,6 +30,12 @@ This module implements all models for Resin Python SDK.
 ## Application
 
 This class implements application model for Resin Python SDK.
+
+Due to the API changes, the returned Application object schema changes as well. Here are the format of old and new returned object.
```
lurch: nits:
 * change `Due to the API changes` to `Due to API changes`
 * change `changes as well` to `has changed`
 * change the last sentence to `Here are the formats of the old and new returned objects.`

EDIT: And obviously if you do make these tweaks, you should also make them in the other places this PR changes too ;-)

DOCUMENTATION.md
```diff
@@ -29,6 +30,12 @@ This module implements all models for Resin Python SDK.
...
+
+The old returned object: `should_track_latest_release, app_name, application, __metadata, actor, git_repository, version, user, device_type, support_expiry_date, commit, id`.
```
lurch: Perhaps `object` here should be changed to `object's properties` ?

lurch:
For reference, this fixes part of #57 

Also pinging @zwalchuk as this is a "documentation" PR ;-)

DOCUMENTATION.md
```diff
@@ -29,6 +30,12 @@ This module implements all models for Resin Python SDK.
...
+
+The new returned object (since Python SDK v1.7.0): `depends_on__application, should_track_latest_release, app_name, __metadata, is_accessible_by_support_until__date, actor, git_repository, version, user, device_type, commit, id`.
```
lurch: Are these listed in any particular order? Would it make sense to list them alphabetically?

nghiant2710:
@lurch updated all your comments, can you review them again?

DOCUMENTATION.md
```diff
@@ -29,6 +30,12 @@ This module implements all models for Resin Python SDK.
 ## Application
 
 This class implements application model for Resin Python SDK.
+
+Due to API changes, the returned Application object schema has changed. Here are the formats of the old and new returned objects.
```
lurch, line 34: Hmmm, seems like these property-lists are only _partially_ alphabetical? (which IMHO is a bit confusing?)
I wrote a mini Python function to check them ;-)
```python
def check_list(old_list):
    new_list = ', '.join(sorted(old_list.split(', ')))
    if new_list != old_list:
        print("List should be:\n%s" % new_list)
```

DOCUMENTATION.md
```diff
@@ -29,6 +30,12 @@ This module implements all models for Resin Python SDK.
...
+
+The old returned object's properties: `__metadata, actor, app_name, application, commit, device_type, git_repository, id, support_expiry_date, user, version, should_track_latest_release`.
```
lurch: List should be:
__metadata, actor, app_name, application, commit, device_type, git_repository, id, should_track_latest_release, support_expiry_date, user, version

DOCUMENTATION.md
```diff
@@ -29,6 +30,12 @@ This module implements all models for Resin Python SDK.
...
+
+The new returned object's properties (since Python SDK v1.7.0): `__metadata, actor, app_name, commit, device_type, git_repository, id, is_accessible_by_support_until__date, should_track_latest_release, user, version, depends_on__application`.
```
lurch: List should be:
__metadata, actor, app_name, commit, depends_on__application, device_type, git_repository, id, is_accessible_by_support_until__date, should_track_latest_release, user, version

DOCUMENTATION.md
```diff
@@ -284,9 +291,73 @@ Revoke support access to an application.
...
+
+The new returned object's properties (since Python SDK v1.7.0): ` __metadata, actor, belongs_to__application, belongs_to__user, created_at, custom_latitude, custom_longitude, device_type, id, ip_address, is_accessible_by_support_until__date, is_connected_to_vpn, is_locked_until__date, is_managed_by__device, is_managed_by__service_instance, is_on__commit, is_online, is_web_accessible, last_connectivity_event, last_vpn_event, latitude, local_id, location, logs_channel, longitude, name, note, os_variant, os_version, provisioning_progress, provisioning_state, public_address, should_be_managed_by__supervisor_release, should_be_running__build, status, supervisor_version, uuid, vpn_address, download_progress`.
```
lurch: List should be:
__metadata, actor, belongs_to__application, belongs_to__user, created_at, custom_latitude, custom_longitude, device_type, download_progress, id, ip_address, is_accessible_by_support_until__date, is_connected_to_vpn, is_locked_until__date, is_managed_by__device, is_managed_by__service_instance, is_on__commit, is_online, is_web_accessible, last_connectivity_event, last_vpn_event, latitude, local_id, location, logs_channel, longitude, name, note, os_variant, os_version, provisioning_progress, provisioning_state, public_address, should_be_managed_by__supervisor_release, should_be_running__build, status, supervisor_version, uuid, vpn_address

lurch (approved):
Couple of minor nits, but otherwise looks good :)

DOCUMENTATION.md
pimterry: This is a breaking change. It looks like we rereleased it as 2.0.0, which is the right thing to do :+1:, but we should also stop people installing 1.7.0, since it'll break their code. Can we not unrelease/deprecate `1.7.0` and change these explanations to reference `2.0.0`?

DOCUMENTATION.md
```diff
@@ -284,9 +291,73 @@ Revoke support access to an application.
...
+    id (str): build id.
+
+#### Returns:
+    dict: build info.
+
+#### Raises:
+    BuildNotFound: if build couldn't be found.
+### Function: get_all_by_application(app_id, include_logs)
+
+Get a specific build.
```
pimterry: ~~This should be 'Get all builds for a specific application'~~ I see this is fixed in https://github.com/resin-io/resin-sdk-python/pull/60

DOCUMENTATION.md
```diff
@@ -284,9 +291,73 @@ Revoke support access to an application.
...
+
+#### Args:
+    app_id (str): application id.
+    include_logs (bool): True if user wants to include build logs in build info, False otherwise.
```
pimterry: Might be good to explain why users should care ('these may be very large')

pimterry (changes requested):

DOCUMENTATION.md
lurch: Hmmm, I suppose in hindsight the "Upgrade Pine endpoint to Resin API v3." of 1.7.0 (see [CHANGELOG](https://github.com/resin-io/resin-sdk-python/blob/master/CHANGELOG.md) ) in #53 should have necessitated a 'major' version-bump rather than a 'minor' version-bump? (it even says `includes many breaking changes` in the PR description)

And in contrast, perhaps adding Python3 support would have been a 'minor' rather than a 'major' change? I guess it depends how strictly you want to apply semver?

DOCUMENTATION.md
lurch: And perhaps it should mention that this parameter can be omitted as it defaults to False? https://github.com/resin-io/resin-sdk-python/blob/master/resin/models/build.py#L46

DOCUMENTATION.md
pimterry: > perhaps adding Python3 support would have been a 'minor' rather than a 'major' change?

I assumed that fixing this was what the 2.0.0 change was for, not adding python 3! Adding python 3 support is almost certainly not a breaking change, you're right.

> I guess it depends how strictly you want to apply semver?

Imo, this isn't a particularly strict interpretation of semver: the currently released version of 1.7.0 is unexpectedly breaking user's code in the real world with no warning (#57), because it wasn't marked as major. There are other cases where it's technically breaking but probably doesn't affect anybody where  it can be pretty debatable, but I think this one is pretty clear cut.

On fixing it though, my suggestion was wrong, the semver spec actually says:

> What do I do if I accidentally release a backwards incompatible change as a minor version?
>
> As soon as you realize that you’ve broken the Semantic Versioning spec, fix the problem and release a new minor version that corrects the problem and restores backwards compatibility. Even under this circumstance, it is unacceptable to modify versioned releases. If it’s appropriate, document the offending version and inform your users of the problem so that they are aware of the offending version.

I.e. we should not remove 1.7.0, but instead ship a 1.8.0 that reverts the API bump just for the 1.x line, to help users easily avoid the breakage (and then continue as normal from 2.0.0). As far as I can tell there's no way to mark a package version as broken or add a warning to it in pypi, but that's what I'd do if this were npm.

@nghiant2710 we should have a chat about the python SDK this week and sort all this out properly. Are you free at 11am UTC tomorrow? (I think that's 6pm your time)

lurch: And of course this also all links (a bit) to #58

DOCUMENTATION.md
nghiant2710: @lurch, it's weird since basically I did the same, the only difference is I used `list.sort()` and print the sorted list instead. From what I understand, `list.sort()` mutates the list while `sorted` returns a new list.

DOCUMENTATION.md
nghiant2710: updated, I marked the parameter as optional one.

DOCUMENTATION.md
nghiant2710: @pimterry, sure 11.am UTC works for me.

pimterry (approved):

DOCUMENTATION.md
```diff
@@ -284,9 +291,73 @@ Revoke support access to an application.
...
+
+#### Raises:
+    BuildNotFound: if build couldn't be found.
+### Function: get_all_by_application(app_id, include_logs)
+
+Get a specific build.
+
+#### Args:
+    app_id (str): application id.
+    include_logs (Optional[bool]): True if user wants to include build logs in build info, False otherwise since these may be very large.
```
lurch: IMHO it'd still be useful for the docs to mention what this optional parameter defaults to.

nghiant2710: @lurch sorry, just forgot to update.

DOCUMENTATION.md
```diff
@@ -284,9 +291,73 @@ Revoke support access to an application.
...
+
+#### Raises:
+    BuildNotFound: if build couldn't be found.
+### Function: get_all_by_application(app_id, include_logs)
+
+Get a specific build.
+
+#### Args:
+    app_id (str): application id.
+    include_logs (Optional[bool]): Defaults to False since these may be very large. True if user wants to include build logs in build info.
```
lurch: :+1:

DOCUMENTATION.md
lurch: Hmmm, I've never played with the API or either of the SDKs, but this function is named `get_all_by_application`, and it's documented as returning a `list of build info`, so it seems weird that the documentation here says `Get a specific build`?

Also, this function is documented as raising `BuildNotFound: if build couldn't be found`, but looking at the code that doesn't seem to be the case? Is there supposed to be an `ApplicationNotFound` or somehing?

nghiant2710: Fixed the description and the exception. It won't return an error even I send request to an application that I don't have access so we can remove the exception here.

DOCUMENTATION.md
```diff
@@ -284,9 +291,70 @@ Revoke support access to an application.
...
+    id (str): build id.
+
+#### Returns:
+    dict: build info.
+
+#### Raises:
+    BuildNotFound: if build couldn't be found.
+### Function: get_all_by_application(app_id, include_logs)
+
+Get list of builds belong to an application.
```
lurch: belong -> belonging

lurch (approved):
Small typo, but otherwise LGTM :)

PR #63, opened 2018-01-18
Automatically update package version in __init__ file

nghiant2710:
connects to https://github.com/resin-io/versionist/pull/100

### DIFF
--- a/versionist.conf.js
+++ b/versionist.conf.js
@@ -20,7 +20,10 @@ module.exports = {
   parseFooterTags: true,
   getGitReferenceFromVersion: 'v-prefix',
   incrementVersion: 'semver',
-  editVersion: false,
+  updateVersion: {
+    preset: 'initPy',
+    targetFile: 'resin/__init__.py'
+  },
 
   // Always add the entry to the top of the Changelog, below the header.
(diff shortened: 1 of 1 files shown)


lurch:
AIUI, we won't be able to merge this PR until the linked versionist feature has been merged, a new version of versionist gets published to npm, VersionBot upgrades to that newer version of versionist, and then that new version of VersionBot gets deployed to k8s (via a keyframe). Is that right @hedss ? :wink:

hedss:
@lurch Yep, that's correct. However, in the short term, the way to do it would be:

* Ensure everything works, get review
* When approved, run new `versionist` on the code, commit changes
* Then set VersionBot to merge. It'll see the changes and will happily merge without trying to update

This does mean you essentially need to merge lock whilst you do it, though.

nghiant2710:
@hedss https://github.com/resin-io/versionist/pull/100 is merged, please let me know when I can proceed

lurch:
@nghiant2710 I think you can proceed if you use versionist >= 2.12.0 ? Unfortunately only versionist 2.11.2 is published to npm so I guess you'll need to install directly from the versionist repo.

PR #65, opened 2018-01-23
Each request times out after 60 seconds

zvin:


jhermsmeier:
Should the timeout be made adjustable, to account for super-slow networks etc.?

zvin:
@jhermsmeier you can always modify the TIMEOUT property of BaseRequest or any child class.

nghiant2710:
@zvin can you please make it a default setting's property here https://github.com/resin-io/resin-sdk-python/blob/master/resin/settings.py#L36-L47? It will be easier for user to modify it.

Also should we make it 30s as the default in JS SDK https://github.com/resin-io/resin-sdk/blob/5ca7650cebe83f14c22b643f62205f26711a2135/lib/models/device.coffee#L49-L52? but I'm not sure about this value, @pimterry might know about this.

zvin:
@nghiant2710 done

ricoder:
This issue was created by me pinging support. Very pleased that the timeout was added so rapidly, any chance of getting this merged into master and a release cut? I am hitting this non-timeout in the field periodically and it is causing data loss.....

resin/base_request.py
```diff
@@ -26,6 +26,10 @@ def __init__(self):
         self.token = Token()
         self.util = Util()
 
+    @property
+    def timeout(self):
+        return self.settings.get('timeout')
```
lurch: Everything in settings is a `str` - do we need to convert this back to an `int` here?

resin/settings.py
```diff
@@ -43,7 +43,9 @@ class Settings(object):
         # cache time : 1 week in milliseconds
         'image_cache_time': str(1 * 1000 * 60 * 60 * 24 * 7),
         # token refresh interval: 1 hour in milliseconds
-        'token_refresh_interval': str(1 * 1000 * 60 * 60)
+        'token_refresh_interval': str(1 * 1000 * 60 * 60),
+        # requests timeout: 30 seconds
+        'timeout': '30',
```
lurch: nit: Is it worth being concerned about the other times in the settings being in milliseconds, but this time being in seconds?

resin/base_request.py
zvin: Good catch, fixed

resin/settings.py
zvin: The requests library takes the timeout parameter in seconds.
I don't mind switching to milliseconds, it's up to you @nghiant2710 .

resin/base_request.py
```diff
@@ -26,6 +26,10 @@ def __init__(self):
         self.token = Token()
         self.util = Util()
 
+    @property
+    def timeout(self):
+        return int(self.settings.get('timeout'))
```
nghiant2710: @zvin we need to convert the timeout to float since requests accepts float for timeout.

resin/settings.py
nghiant2710: Makes sense. Please switch to milliseconds @zvin

resin/base_request.py
zvin: Updated

resin/settings.py
```diff
@@ -43,7 +43,9 @@ class Settings(object):
         # cache time : 1 week in milliseconds
         'image_cache_time': str(1 * 1000 * 60 * 60 * 24 * 7),
         # token refresh interval: 1 hour in milliseconds
-        'token_refresh_interval': str(1 * 1000 * 60 * 60)
+        'token_refresh_interval': str(1 * 1000 * 60 * 60),
+        # requests timeout: 30 seconds in milliseconds
+        'timeout': str(1000 * 30),
```
lurch: super-small nit: could we switch this round to `str(30 * 1000)` so that it matches the other settings?

zvin: done

nghiant2710 (approved):

lurch (approved):

lurch:
Hmmm, should this fix also be made to the 1.x branch as well as the 2.x branch? @pimterry

pimterry:
@lurch Imo no, we don't need this for 1.x.

We do breaking changes so that users can quickly interpret (and potentially automatically start using) new versions as they come out. We're not using version numbers to demarcate different supported versions you should use in different scenarios.

Right now afaik we're always only actively supporting one head version on any given project, so once a change gets merged (breaking or otherwise), old versions are effectively frozen. Trying to do anything else gets much more complicated, and unless there's a really good reason (like a critical security fix), I'd much rather move forwards with new changes always only on top of the latest version, and encourage users to upgrade.

lurch:
Cool, thanks for the clarification @pimterry :+1:

Issue #39, opened 2016-09-03
Create test suite for Python SDK


nghiant2710:

otaviojacobi:
Several new tests were added as #326 and the test suite was already created previously. Closing this in favor of a new issue to increase the coverage.

PR #67, opened 2018-01-31
Test suite for Python SDK (Dont merge this)

nghiant2710:
This is how we organize the test suite for Python SDK (there are only test cases for `application` class atm but more will be added soon).

This will fix a part of https://github.com/resin-io/resin-sdk-python/issues/39.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "2.6"
   - "2.7"
   - "3.5"
   - "3.6"
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,11 @@ python:
   - "3.5"
   - "3.6"
 # command to install dependencies
-install: "pip install pep8"
-# command to run tests
-script: python ./tests/pep8-git-pr-checker
+install: 
+- pip install pep8 setuptools
+- python setup.py install
+script:
+# Stye Check 
+- python ./tests/pep8-git-pr-checker
+# Automated Testing
+- python -m unittest discover tests -v
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ install:
 - pip install pep8 setuptools
 - python setup.py install
 script:
-# Stye Check 
+# Style Check 
 - python ./tests/pep8-git-pr-checker
 # Automated Testing
 - python -m unittest discover tests -v
--- a/resin/base_request.py
+++ b/resin/base_request.py
@@ -46,17 +46,17 @@ class BaseRequest(object):
     def __post(self, url, headers, data, stream=None):
         self.__set_content_type(headers, 'application/json')
         if not stream:
-            return requests.post(url, data=json.dumps(self.util.to_utf(data)), headers=headers, timeout=self.timeout)
+            return requests.post(url, data=json.dumps(self.util.decode_utf8(data)), headers=headers, timeout=self.timeout)
         return requests.post(
-            url, data=json.dumps(self.util.to_utf(data)), headers=headers, stream=stream, timeout=self.timeout)
+            url, data=json.dumps(self.util.decode_utf8(data)), headers=headers, stream=stream, timeout=self.timeout)
 
     def __put(self, url, headers, data=None, stream=None):
         self.__set_content_type(headers, 'application/json')
-        return requests.put(url, data=json.dumps(self.util.to_utf(data)), headers=headers, timeout=self.timeout)
+        return requests.put(url, data=json.dumps(self.util.decode_utf8(data)), headers=headers, timeout=self.timeout)
 
     def __patch(self, url, headers, data=None, stream=None):
         self.__set_content_type(headers, 'application/json')
-        return requests.patch(url, data=json.dumps(self.util.to_utf(data)), headers=headers, timeout=self.timeout)
+        return requests.patch(url, data=json.dumps(self.util.decode_utf8(data)), headers=headers, timeout=self.timeout)
 
     def __delete(self, url, headers, data=None, stream=None):
         self.__set_content_type(headers, 'application/x-www-form-urlencoded')
@@ -200,7 +200,7 @@ class Util(object):
         # return None if key is not present
         return os.environ.get('RESIN_API_KEY')
 
-    def to_utf(self, source):
+    def decode_utf8(self, source):
         return {
             (k.decode('utf-8') if type(k).__name__ == 'bytes' else k):
             (v.decode('utf-8') if type(v).__name__ == 'bytes' else v) for k, v in source.items()
--- a/resin/base_request.py
+++ b/resin/base_request.py
@@ -46,17 +46,17 @@ class BaseRequest(object):
     def __post(self, url, headers, data, stream=None):
         self.__set_content_type(headers, 'application/json')
         if not stream:
-            return requests.post(url, data=json.dumps(data), headers=headers, timeout=self.timeout)
+            return requests.post(url, data=json.dumps(self.util.to_utf(data)), headers=headers, timeout=self.timeout)
         return requests.post(
-            url, data=json.dumps(data), headers=headers, stream=stream, timeout=self.timeout)
+            url, data=json.dumps(self.util.to_utf(data)), headers=headers, stream=stream, timeout=self.timeout)
 
     def __put(self, url, headers, data=None, stream=None):
         self.__set_content_type(headers, 'application/json')
-        return requests.put(url, data=json.dumps(data), headers=headers, timeout=self.timeout)
+        return requests.put(url, data=json.dumps(self.util.to_utf(data)), headers=headers, timeout=self.timeout)
 
     def __patch(self, url, headers, data=None, stream=None):
         self.__set_content_type(headers, 'application/json')
-        return requests.patch(url, data=json.dumps(data), headers=headers, timeout=self.timeout)
+        return requests.patch(url, data=json.dumps(self.util.to_utf(data)), headers=headers, timeout=self.timeout)
 
     def __delete(self, url, headers, data=None, stream=None):
         self.__set_content_type(headers, 'application/x-www-form-urlencoded')
@@ -199,3 +199,9 @@ class Util(object):
     def get_api_key(self):
         # return None if key is not present
         return os.environ.get('RESIN_API_KEY')
+
+    def to_utf(self, source):
+        return {
+            (k.decode('utf-8') if type(k).__name__ == 'bytes' else k):
+            (v.decode('utf-8') if type(v).__name__ == 'bytes' else v) for k, v in source.items()
+        }
--- a/tests/functional/models/test_application.py
+++ b/tests/functional/models/test_application.py
@@ -1,0 +1,145 @@
+import unittest
+import json
+from datetime import datetime
+
+from tests.helper import TestHelper
+
+
+class TestApplication(unittest.TestCase):
+
+    helper = None
+    resin = None
+
+    @classmethod
+    def setUpClass(cls):
+        cls.helper = TestHelper()
+        cls.resin = cls.helper.resin
+
+    def tearDown(self):
+        # Wipe all apps after every test case.
+        self.helper.wipe_application()
+
+    def test_create(self):
+        # should be rejected if the device type is invalid
+        with self.assertRaises(self.helper.resin_exceptions.InvalidDeviceType):
+            self.resin.models.application.create('FooBar', 'Foo')
+
+        # should be rejected if the name has less than three characters
+        with self.assertRaises(Exception) as cm:
+            self.resin.models.application.create('Fo', 'Raspberry Pi 2')
+        self.assertIn('It is necessary that each app name that is of a user (Auth), has a Length (Type) that is greater than or equal to 4', cm.exception.message)
+
+        # should be able to create an application
+        app = json.loads(self.resin.models.application.create('FooBar', 'Raspberry Pi 2').decode('utf-8'))
+        self.assertEqual(app['app_name'], 'FooBar')
+
+    def test_get_all(self):
+        # given no applications, it should return empty list.
+        self.assertEqual(self.resin.models.application.get_all(), [])
+
+        # given there is an application, it should return a list with length 1.
+        self.resin.models.application.create('FooBar', 'Raspberry Pi 2')
+        self.assertEqual(len(self.resin.models.application.get_all()), 1)
+
+    def test_get(self):
+        # raise resin.exceptions.ApplicationNotFound if no application found.
+        with self.assertRaises(self.helper.resin_exceptions.ApplicationNotFound):
+            self.resin.models.application.get('AppNotExist')
+
+        # found an application, it should return an application with matched name.
+        self.resin.models.application.create('FooBar', 'Raspberry Pi 2')
+        self.assertEqual(self.resin.models.application.get('FooBar')['app_name'], 'FooBar')
+
+    def test_has(self):
+        # should be true if the application name exists, otherwise it should return false.
+        self.resin.models.application.create('FooBar', 'Raspberry Pi 2')
+        self.assertFalse(self.resin.models.application.has('FooBar1'))
+        self.assertTrue(self.resin.models.application.has('FooBar'))
+
+    def test_has_any(self):
+        # given no applications, it should return false.
+        self.assertFalse(self.resin.models.application.has_any())
+
+        # should return true if at least one application exists.
+        self.resin.models.application.create('FooBar', 'Raspberry Pi 2')
+        self.assertTrue(self.resin.models.application.has_any())
+
+    def test_get_by_id(self):
+        # raise resin.exceptions.ApplicationNotFound if no application found.
+        with self.assertRaises(self.helper.resin_exceptions.ApplicationNotFound):
+            self.resin.models.application.get_by_id(1)
+
+        # found an application, it should return an application with matched id.
+        app = json.loads(self.resin.models.application.create('FooBar', 'Raspberry Pi 2').decode('utf-8'))
+        self.assertEqual(self.resin.models.application.get_by_id(app['id'])['id'], app['id'])
+
+    def test_remove(self):
+        # should be able to remove an existing application by name.
+        app = json.loads(self.resin.models.application.create('FooBar', 'Raspberry Pi 2').decode('utf-8'))
+        self.assertEqual(len(self.resin.models.application.get_all()), 1)
+        self.resin.models.application.remove('FooBar')
+        self.assertEqual(len(self.resin.models.application.get_all()), 0)
+
+    def test_generate_provisioning_key(self):
+        # should be rejected if the application id does not exist.
+        with self.assertRaises(self.helper.resin_exceptions.ApplicationNotFound):
+            self.resin.models.application.generate_provisioning_key('5685')
+
+        # should be able to generate a provisioning key by app id.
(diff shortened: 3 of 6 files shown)


lurch:
@pimterry Does dropping support for Python 2.6 count as a "breaking" (and hence major) change?

resin/base_request.py
```diff
@@ -199,3 +199,9 @@ def should_update_token(self, age, token_fresh_interval):
     def get_api_key(self):
         # return None if key is not present
         return os.environ.get('RESIN_API_KEY')
+
+    def to_utf(self, source):
```
lurch: Should this be named something like `decode_utf8` rather than `to_utf` ?

resin/base_request.py
```diff
@@ -199,3 +199,9 @@ def should_update_token(self, age, token_fresh_interval):
...
+        return {
+            (k.decode('utf-8') if type(k).__name__ == 'bytes' else k):
+            (v.decode('utf-8') if type(v).__name__ == 'bytes' else v) for k, v in source.items()
```
lurch: Just out of curiosity, where are these `'bytes'` values coming from?

tests/functional/models/test_application.py
```diff
@@ -0,0 +1,145 @@
...
+        self.resin.models.application.create('FooBar', 'Raspberry Pi 2')
+        self.assertTrue(self.resin.models.application.has_any())
+
+    def test_get_by_id(self):
+        # raise resin.exceptions.ApplicationNotFound if no application found.
+        with self.assertRaises(self.helper.resin_exceptions.ApplicationNotFound):
+            self.resin.models.application.get_by_id(1)
+
+        # found an application, it should return an application with matched id.
+        app = json.loads(self.resin.models.application.create('FooBar', 'Raspberry Pi 2').decode('utf-8'))
```
lurch, line 76: Is having to do `json.loads(some_api_call().decode('utf-8'))` a "code smell" that `some_api_call()` isn't operating at the right level of abstraction?  (i.e. would it make more sense for the `decode('utf-8')` and `json.loads()` to be done _inside_ `some_api_call()`? Although this would obviously be a breaking change)

tests/functional/models/test_application.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def test_create(self):
+        # should be rejected if the device type is invalid
+        with self.assertRaises(self.helper.resin_exceptions.InvalidDeviceType):
+            self.resin.models.application.create('FooBar', 'Foo')
+
+        # should be rejected if the name has less than three characters
+        with self.assertRaises(Exception) as cm:
+            self.resin.models.application.create('Fo', 'Raspberry Pi 2')
+        self.assertIn('It is necessary that each app name that is of a user (Auth), has a Length (Type) that is greater than or equal to 4', cm.exception.message)
```
lurch: Curiously, the comment a few lines above claims that the application name has to be >= 3 characters, and yet the error message that you're checking implies that the application name has has to be >= 4 characters? They can't both be correct ;)

tests/functional/models/test_application.py
```diff
@@ -0,0 +1,145 @@
...
+        with self.assertRaises(self.helper.resin_exceptions.ApplicationNotFound):
+            self.resin.models.application.get_by_id(1)
+
+        # found an application, it should return an application with matched id.
+        app = json.loads(self.resin.models.application.create('FooBar', 'Raspberry Pi 2').decode('utf-8'))
+        self.assertEqual(self.resin.models.application.get_by_id(app['id'])['id'], app['id'])
+
+    def test_remove(self):
+        # should be able to remove an existing application by name.
+        app = json.loads(self.resin.models.application.create('FooBar', 'Raspberry Pi 2').decode('utf-8'))
```
lurch, line 81: This `app = json.loads(self.resin.models.application.create('FooBar', 'Raspberry Pi 2').decode('utf-8'))` is being done so many times, that perhaps it's worth moving it to e.g. a `create_test_application(self)` function?

EDIT: e.g.
```python
def create_test_application(self, name='FooBar', device_type='Raspberry Pi 2'):
    return json.loads(self.resin.models.application.create(name, device_type).decode('utf-8'))
```
as that'd allow you to just use `app = self.create_test_application()` in most places, while still allowing the `name` and `device_type` to be specified as needed in other tests.

tests/functional/models/test_application.py
```diff
@@ -0,0 +1,145 @@
...
+        self.assertEqual(len(key), 32)
+
+    def test_enable_rolling_updates(self):
+        # should enable rolling update for the applications devices.
+        app = json.loads(self.resin.models.application.create('FooBar', 'Raspberry Pi 2').decode('utf-8'))
+        self.resin.models.application.enable_rolling_updates(app['id'])
+        app = self.resin.models.application.get('FooBar')
+        self.assertTrue(app['should_track_latest_release'])
+
+    def test_disable_rolling_updates(self):
```
lurch, line 104: With all these enable and disable tests, I wonder if it _might_ make more sense to do e.g.
```python
    def test_rolling_updates(self): # or test_toggle_rolling_updates
        app = json.loads(self.resin.models.application.create('FooBar', 'Raspberry Pi 2').decode('utf-8'))
        # I guess at this point you could also test the default value of app['should_track_latest_release'] ?
        # should enable rolling update for the applications devices.
        self.resin.models.application.enable_rolling_updates(app['id'])
        app = self.resin.models.application.get('FooBar')
        self.assertTrue(app['should_track_latest_release'])
        # should disable rolling update for the applications devices.
        self.resin.models.application.disable_rolling_updates(app['id'])
        app = self.resin.models.application.get('FooBar')
        self.assertFalse(app['should_track_latest_release'])
```
?
And if the default value of `app['should_track_latest_release']` is True, then it'd probably make sense to test disabling it before testing enabling it.

tests/functional/models/test_application.py
```diff
@@ -0,0 +1,145 @@
...
+        # should throw an error if the expiry timestamp is in the past.
+        expiry_timestamp = int((datetime.utcnow() - datetime.utcfromtimestamp(0)).total_seconds() * 1000 - 10000)
+        with self.assertRaises(self.helper.resin_exceptions.InvalidParameter):
+            self.resin.models.application.grant_support_access(app['id'], expiry_timestamp)
+
+        # should grant support access until the specified time.
+        expiry_time = int((datetime.utcnow() - datetime.utcfromtimestamp(0)).total_seconds() * 1000 + 3600 * 1000)
+        self.resin.models.application.grant_support_access(app['id'], expiry_time)
+        support_date = datetime.strptime(self.resin.models.application.get('FooBar')['is_accessible_by_support_until__date'], '%Y-%m-%dT%H:%M:%S.%fZ')
+        self.assertEqual(int((support_date - datetime.utcfromtimestamp(0)).total_seconds() * 1000), expiry_time)
```
lurch: Might make sense to tweak this to:
```python
        support_expiry_time = int((support_date - datetime.utcfromtimestamp(0)).total_seconds() * 1000)
        self.assertEqual(support_expiry_time, expiry_time)
```
(IMHO it makes it slightly more obvious what's going on, and this'll probably be easier to debug should any problems arise)

tests/helper.py
```diff
@@ -0,0 +1,52 @@
+from resin import Resin
+from resin import exceptions as resin_exceptions
+from resin import Token
+
+
+class TestHelper(object):
+
+    credential = {
+        'email': 'pythonsdktest@gmail.com',
+        'password': 'resintest',
```
lurch: @pimterry Do we really want to be storing resin account credentials in plaintext? How are other API / SDK tests done?
(I guess anybody could log in with the details above, change the account's password, and then suddenly all the PythonSDK CI tests would break)

tests/helper.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        Wipe all user's apps
+        """
+
+        self.resin.models.application.base_request.request(
+            'application', 'DELETE',
+            endpoint=self.resin.settings.get('pine_endpoint'), login=True
+        )
+
+    def reset_user(self):
```
lurch, line 75: This isn't used anywhere (yet) ?

tests/functional/models/test_application.py
```diff
@@ -0,0 +1,145 @@
...
+        app = json.loads(self.resin.models.application.create('FooBar', 'Raspberry Pi 2').decode('utf-8'))
+        self.assertEqual(app['app_name'], 'FooBar')
+
+    def test_get_all(self):
+        # given no applications, it should return empty list.
+        self.assertEqual(self.resin.models.application.get_all(), [])
+
+        # given there is an application, it should return a list with length 1.
+        self.resin.models.application.create('FooBar', 'Raspberry Pi 2')
+        self.assertEqual(len(self.resin.models.application.get_all()), 1)
```
lurch: As there's only one application, you could also check the `app_name`, as you do in the next test?

Would it also be worth creating a test for adding two applications, to check that `get_all()` can cope with more than one result? ;) And you could check e.g.
```python
all_applications = self.resin.models.application.get_all()
self.assertEqual(len(all_applications), 2)
self.assertNotEqual(all_applications[0]['app_name'], all_applications[1]['app_name'])
```

tests/functional/models/test_application.py
```diff
@@ -0,0 +1,145 @@
...
+        # Wipe all apps after every test case.
+        self.helper.wipe_application()
+
+    def test_create(self):
+        # should be rejected if the device type is invalid
+        with self.assertRaises(self.helper.resin_exceptions.InvalidDeviceType):
+            self.resin.models.application.create('FooBar', 'Foo')
+
+        # should be rejected if the name has less than three characters
+        with self.assertRaises(Exception) as cm:
```
lurch: Is this just `Exception` and not some more specific type?

tests/functional/models/test_application.py
lurch: Or in fact, perhaps adding
```python
def datetime_to_epoch_ms(datevalue):
    return int((datevalue - datetime.utcfromtimestamp(0)).total_seconds() * 1000)
```
would be useful, as there are several places in these unit-tests where that could be used :-)

tests/functional/models/test_application.py
lurch: In fact, this is actually an inconsistency in the docs as https://github.com/resin-io/resin-sdk-python/blob/master/DOCUMENTATION.md#function-createname-device_type says "Returns: dict: application info." but then the example shows it just returning a single string. Should I create a separate issue for this?

lurch (commented):
Left 'a few' comments, but this looks like a great start to the test suite :+1:

tests/functional/models/test_application.py
lurch: Haha, which is also what https://stackoverflow.com/a/11111177 says ;)

.travis.yml
```diff
@@ -1,10 +1,14 @@
...
   - "3.6"
 # command to install dependencies
-install: "pip install pep8"
-# command to run tests
-script: python ./tests/pep8-git-pr-checker
+install: 
+- pip install pep8 setuptools
+- python setup.py install
+script:
+# Stye Check
```
pimterry: Nit: Stye -> Style

tests/functional/models/test_application.py
pimterry: This is definitely a bit awkward, and good to fix, though it's fine to do in a later PR

tests/functional/models/test_application.py
pimterry: Personally I'd rather have more & smaller tests - makes them easier to read, easier to tell what exactly is broken if they fail, and easier to run separate bits in isolation.

tests/helper.py
pimterry: Indeed not. The SDK credentials read from the environment, loading a `.env` file automatically if present, so you just create a `.env` with credentials to run tests locally, and we set the env vars in the Travis/etc build config. We should do that here too (and reset these login details), both for security, and because that way you can use different login details locally/in CI. If you use the same login details everywhere then running the tests locally will break any running CI builds.

tests/helper.py
```diff
@@ -0,0 +1,52 @@
...
+                **{
+                    'username': self.credential['user_id'],
+                    'password': self.credential['password']
+                }
+            )
+
+        # Stop the test if it's run by an admin user account.
+        token = Token()
+        if any('admin' in s for s in token.get_data()['permissions']):
+            raise Exception('The test is run with an admin user account. Cancelled, please try again with a normal account!')
```
pimterry, line 34: :+1: :+1: :+1: Good move! I need to do this for the JS tests too.

tests/helper.py
```diff
@@ -0,0 +1,52 @@
...
+            'application', 'DELETE',
+            endpoint=self.resin.settings.get('pine_endpoint'), login=True
+        )
+
+    def reset_user(self):
+        """
+        Wipe all user's apps and ssh keys added.
+        """
+
+        print(self.resin.auth.who_am_i())
```
pimterry: This looks like a debugging leftover, should probably be removed

tests/helper.py
pimterry: Yep. It should be - you probably just want a single `reset_all` method that includes both, and to then run that from the tests instead.

pimterry (changes requested):
A few comments, but this looks like a great start. The one main thing we need to resolve I think is getting credentials from the env instead of hardcoding them.

The point about breaking changes is an interesting one - if this PR is actually removing support for Python 2.6 somehow, this is a breaking change. If we're stopping advertising it but it will keep working at least until the next major bump, or if it's actually never worked at all for the current major version and it's a documentation bug, then it isn't.

tests/functional/models/test_application.py
lurch: Fair enough. Should there also be tests for the default values? And presumably if the default value is e.g. True, then when testing the "enable" function you'd first need to disable it, to ensure that the enable isn't just a NOP?

tests/functional/models/test_application.py
nghiant2710: Created an issue for this https://github.com/resin-io/resin-sdk-python/issues/68. I couldn't remember why I made it like that but we need to fix this.

tests/functional/models/test_application.py
nghiant2710: As mentioned above, I will fix the return value for this function so will update the test cases in the PR for this issue https://github.com/resin-io/resin-sdk-python/issues/68 later.

tests/functional/models/test_application.py
nghiant2710: @lurch, I don't think we need a test for default values here but it makes sense to enable/disable first before testing disable/enable function to ensure our test is doing a right thing.

tests/functional/models/test_application.py
nghiant2710: 👍 Updated

tests/helper.py
nghiant2710: Fixed. I've created another PR just to see the test result https://github.com/resin-io/resin-sdk-python/pull/69 since travis doesn't allow PR from downstream repo to access encrypted variables.

tests/helper.py
nghiant2710: @lurch @pimterry since these teardown functions are run after every test case so I want to keep it as small as possible, really don't want to spam the API with unnecessary requests. The fully reset function isn't used yet but sure we will use it for coming test case.

tests/helper.py
nghiant2710: Updated

nghiant2710:
@lurch @pimterry, I know dropping support for Python 2.6 is a breaking change but on the other hand, I'm pretty sure that no one is using the SDK with Python 2.6 so this change won't break anyone application. Moreover, we want the test suite landed as soon as possible (Don't want to wait until next major version is merged) so I make it a `minor` change.

Anyway, I don't have a strong opinion about this so if we agree this is a breaking change then I will try to finish the while test suite soon and then bump major version.

resin/base_request.py
nghiant2710: updated

resin/base_request.py
nghiant2710: In Python 3, a bytes string can't be dumped into JSON object so we need to encode any bytes strings using the encoding `utf-8` first.

tests/functional/models/test_application.py
nghiant2710: Yeah, we want to validate the error message here so a general exception is fine with me.

tests/functional/models/test_application.py
nghiant2710: Fixed. I copied from the JS test so we probably need to fix it here as well https://github.com/resin-io/resin-sdk/blob/master/tests/integration/models/application.spec.coffee#L59-L64

pimterry:
@nghiant2710 if people are using this in 2.6, would this PR break their code?

nghiant2710:
@pimterry, it won't break their code, just the test suite won't be able to run on Python 2.6

tests/functional/models/test_application.py
nghiant2710: added this to the test case.

pimterry:
@nghiant2710 I think that's totally fine then, and this is just a minor change.

As long as their code still works, everything's ok, and all we're really doing with this PR is changing which version support we explicitly advertise.

tests/helper.py
```diff
@@ -28,6 +33,35 @@ def __init__(self):
...
+                raise Exception('Mandatory env keys missing!')
+            cls.credentials = config_data
+        else:
+            # If .env file not exists, read credentials from environment vars.
+            try:
+                cls.credentials['email'] = os.environ.get('TEST_ENV_EMAIL')
+                cls.credentials['user_id'] = os.environ.get('TEST_ENV_USER_ID')
+                cls.credentials['password'] = os.environ.get('TEST_ENV_PASSWORD')
+            except:
+                raise Exception('Mandatory env keys missing!')
```
pimterry: This will never be thrown, as `os.environ.get` doesn't throw if the variable doesn't exist (it just returns `None`)

pimterry (changes requested):
One last small bug in the credential loading. The rest all looks good to me.

tests/helper.py
nghiant2710: @pimterry updated, by doing `os.environ['Key']`, it will raise `KeyError` exception, if the environment variable is not present.

pimterry (approved):
All the code looks good to me now. The build is failing though because env vars aren't provided to PRs from external sources, only PRs on the repo itself. It looks like you've submitted this from a PR on your own branch?

That's unusual - I think the standard Resin process is to open branches directly on resin repo itself, and PR from there. You'll need to move to that approach to make this work. Otherwise from Github's point of view you're an unknown external contributor, so you shouldn't be given free access to the credentials.

That'll mean reopening this PR, but it's otherwise easy: push this same branch directly to this repo instead, open a new PR with that, and I'll approve so you can merge.

nghiant2710:
@pimterry, I noticed that travis won't include encrypted env vars if the PR is from external sources, that's why i created this PR to show the test result https://github.com/resin-io/resin-sdk-python/pull/69.

From now on, I will create PR from branches belong to resin repo.

nghiant2710:
@hedss, I already pushed the v2.1.1 release tags, do you know why VersionBot still reports this error?

lurch:
I think for VB to be able to do the merging, the PR needs to be coming from the same repo and not from the "external" `nghiant2710/resin-sdk-python` repo? Might be easiest to close this PR, and reapply the commits on a branch from `resin-io/resin-sdk-python` ?

Issue #68, opened 2018-02-07
`resin.models.application.create()` returns incorrect object


nghiant2710:
It currently returns a string but should return a dict instead.

PR #70, opened 2018-02-09
 Fix return object of `create` functions to ensure they return a dict.

nghiant2710:
Fixes https://github.com/resin-io/resin-sdk-python/issues/68
Change-Type: major

### DIFF
--- a/resin/models/device.py
+++ b/resin/models/device.py
@@ -2,6 +2,7 @@ import sys
 import binascii
 import os
 from datetime import datetime
+import json
 
 try:  # Python 3 imports
     from urllib.parse import urljoin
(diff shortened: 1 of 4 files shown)


lurch:
Looks like there's still some inline-documentation that needs fixing?
* https://github.com/resin-io/resin-sdk-python/pull/70/files#diff-7c5331951da4fb6c031e21602cdbeaf4R171 shows a string being returned
* https://github.com/resin-io/resin-sdk-python/pull/70/files#diff-bbefe39d0e484a0b896b8c5172208ffcR539 says it returns a`str` rather than a `dict`
* https://github.com/resin-io/resin-sdk-python/pull/70/files#diff-bbefe39d0e484a0b896b8c5172208ffcR544 shows a string being returned
* https://github.com/resin-io/resin-sdk-python/pull/70/files#diff-47864fcd322f9ddb6d8527bf34475418R235 I _think_ the outer-most quotes should be removed?

And if you could regenerate the README too, that'd be nice ;-)

.gitignore
```diff
@@ -1,3 +1,4 @@
 *.pyc
 test.py
 *.py.bak
+.env*
```
lurch: Was this change _supposed_ to be included in this PR?

nghiant2710: yeah, it shouldn't be included here.

nghiant2710:
@lurch, I've updated your comments. Can you please check it again?

resin/models/device.py
```diff
@@ -531,12 +532,12 @@ def register(self, app_id, uuid):
             uuid (str): device uuid.
 
         Returns:
-            str: dictionary contains device info (can be parsed to dict).
+            dict: dictionary contains device info (can be parsed to dict).
```
lurch: You can get rid of the " (can be parsed to dict)" here now :)

lurch (commented):
For the purposes of consistency, would it be worth changing all the example return-dicts to use single-quotes? (This obviously doesn't really matter though)

resin/models/device.py
nghiant2710: Updated!

I fixed the single-quotes comment as well.

pimterry (approved):

resin/models/environment_variables.py
```diff
@@ -228,11 +228,11 @@ def create(self, app_id, env_var_name, value):
             value (str): environment variable value.
 
         Returns:
-            str: new application environment info.
+            dict: new application environment info.
```
lurch, line 231: Small nit, but should this say "application environment variable info" instead of "application environment info".

lurch (approved):
Small nit, but looks good :+1: 

Out of curiosity, how come the URLs in the example returned dictionaries sometimes start with `/ewa/...` and sometimes start with `/resin/...`?

PR #75, opened 2018-03-19
Python SDK v3

nghiant2710:
This PR contains some fixes for API v3. We need to release this before adding API v4 (multicontainer) support.

Most of the commits are reviewed, only the last one is added recently.


resin/models/build.py
```diff
@@ -43,6 +43,25 @@ def get(self, id):
         else:
             raise exceptions.BuildNotFound(id)
 
+    def get_by_commit(self, commit_hash):
+        """
+        Get a specific build by commit hash.
```
pimterry: I think the wording isn't quite right: there can easily be multiple builds for a commit hash (just only one successful build), so this should be something like `Get all builds for a commit hash`

resin/models/device.py
```diff
@@ -692,34 +695,46 @@ def disable_device_url(self, uuid):
...
 
         Args:
             uuid (str): device uuid.
-            build (str): build id.
+            build_commit_hash (str): build commit hash.
 
         Raises:
             DeviceNotFound: if device couldn't be found.
             ApplicationNotFound: if application couldn't be found.
+            FailedBuild: if build commit hash points to an error build.
```
pimterry: For similar reasons, this should be `if all builds for the commit hash failed.`

pimterry (changes requested):

resin/models/build.py
nghiant2710: @pimterry updated!

resin/models/device.py
nghiant2710: Updated!

pimterry (approved):

PR #76, opened 2018-03-19
Python SDK v4

nghiant2710:
This PR adds support for API v4 and  bases on top of Python SDK v3 (https://github.com/resin-io/resin-sdk-python/pull/75) so DO NOT MERGE IT before the SDK v3 PR is merged.


resin/models/config_variable.py
```diff
@@ -0,0 +1,262 @@
...
+
+        params = {
+            'filter': 'device',
+            'eq': device['id']
+        }
+
+        return self.base_request.request(
+            'device_config_variable', 'GET', params=params,
+            endpoint=self.settings.get('pine_endpoint')
+        )['d']
```
pimterry, line 62: This does two requests to get the vars - that works, but it is possible to do a pine query that gets all device_config_variables from a UUID directly, in just one request. Not sure how concerned we are about the performance implications here, but might be worth looking at (same applies for many of the below too)

resin/models/config_variable.py
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        return json.loads(self.base_request.request(
+            'application_config_variable', 'POST', data=data,
+            endpoint=self.settings.get('pine_endpoint')
+        ).decode('utf-8'))
+
+    def update(self, var_id, value):
+        """
+        Update am application config variable.
```
pimterry: `am` -> `an`

resin/models/environment_variables.py
```diff
@@ -61,57 +54,73 @@ def get_all(self, uuid):
...
+            params = {
+                'filter': 'service_install',
+                'eq': service_install['id']
+            }
+
+            env_vars += (self.base_request.request(
+                'device_service_environment_variable', 'GET', params=params,
+                endpoint=self.settings.get('pine_endpoint')
+            )['d'])
+        return env_vars
```
pimterry: This is doing N + 1 requests to get environment variables (where N is the number of services on your device). That's going to be _really_ slow if you have more than a couple of services, and is going to hammer the API. You can do this in one pine query, and we definitely should.

The UI does this one query for example with this: https://github.com/resin-io/resin-ui/blob/250de0c375e28fc2b5c0237ecd092620aefd69a1/src/services/models/environment-model-service.ts#L121-L147

resin/models/environment_variables.py
```diff
@@ -146,118 +155,100 @@ def remove(self, var_id):
...
+        for service in services:
+            params = {
+                'filter': 'service',
+                'eq': service['id']
+            }
+
+            env_vars += self.base_request.request(
+                'service_environment_variable', 'GET', params=params,
+                endpoint=self.settings.get('pine_endpoint')
+            )['d']
```
pimterry: Ditto

resin/models/release.py
```diff
@@ -0,0 +1,113 @@
...
+        release = self.get(id)
+
+        # Details of images built, including image id and service name
+        for service in release['composition']['services']:
+            images.append({
+                'service_name': service,
+                'id': self.image._Image__get_by_option(
+                    'is_a_build_of__service',
+                    self.service._Service__get_by_option('service_name', service)[0]['id']
+                )['id']
```
pimterry: This is similar - it's doing a sequential request for each service, and it looks like it's parsing the composition itself to work out what services exist.

You don't need to do that, and we shouldn't: instead, you can do a single request that will return you the release along with all its associated images. See https://github.com/resin-io/resin-sdk/blob/multicontainer/lib/models/release.coffee#L104-L117 for the JS SDK equivalent.

pimterry (changes requested):

nghiant2710:
@pimterry, about those requests, I have the same feeling as you but we don't have a pine client for Python or at least an OData query parser for now (I thought about those features in the past but never got a chance to implement them) This will be my first priority for the coming PR because we have more complicated requests since API v4. For this PR, I just want to deliver an initial support for API v4 as quick as possible since it's indeed a breaking change as users won't be able to interact with MC apps without these updates.

pimterry:
For the functions making two requests instead of one, I don't mind merging and fixing later.

I really don't think it's a good idea to release SDK methods that make N+1 sequential API calls though. It's going to be extremely slow for a lot users, and it'll hammer the API if anybody tries to use it for anything substantial.

There's a couple of easy options imo:

* Pull out those N+1 methods (`Release.get_with_image_details`, `ServiceEnvVariable.get_all` and `DeviceServiceEnvVariable.get_all`) for now, ship the rest, and fix & add those later
* Move to the new query format by just building the query as string. Should be easy enough to get a working example using the dashboard and looking at the requests it makes, and we should be able to use that immediately without a real pine client.

I'm curious what @Page- thinks, since he had strong opinions on similar issues in the JS world. Or it might be something for an arch call? It depends on our priorities here on doing this nicely, vs shooting a Python SDK out the door ASAP no matter what.

nghiant2710:
@pimterry, I've implemented the second option which seems to be the best choice for now.

pimterry (approved):

PR #80, opened 2018-04-09
Add support for tag models

nghiant2710:


docs_generator.py
```diff
@@ -26,6 +26,9 @@
     ('Image', TOC_L2),
     ('Release', TOC_L2),
     ('Service', TOC_L2),
+    ('Tag', TOC_L2),
+    ('DeviceTag', TOC_L3),
+    ('ApplicationTag', TOC_L3),
```
thgreasi: How about moving `ApplicationTag` before `DeviceTag`?

resin/models/tag.py
```diff
@@ -0,0 +1,308 @@
...
+            app_id (str): application id.
+            tag_key (str): tag key.
+            value (str): tag value.
+
+        Returns:
+            dict: dict contains application tag info if tag doesn't exist.
+            OK: if tag exists.
+
+        Examples:
+            >>> resin.models.tag.application.set('1005767', 'tag1','Python SDK')
```
thgreasi: NIT: add a space between `'tag1','Python SDK'`

resin/models/tag.py
```diff
@@ -0,0 +1,308 @@
...
+    def remove(self, app_id, tag_key):
+        """
+        Remove an application tag.
+
+        Args:
+            app_id (str): application id.
+            tag_key (str): tag key.
+
+        Examples:
+            >>> resin.models.tag.application.set('1005767', 'tag1')
```
thgreasi: The example here should be updated to demonstrate `remove`.

resin/models/tag.py
```diff
@@ -0,0 +1,308 @@
...
+    This class is a wrapper for Tag models.
+
+    """
+
+    def __init__(self):
+        self.device = DeviceTag()
+        self.application = ApplicationTag()
+
+
+class BaseTag(object):
```
thgreasi, line 20: Since we are going to be adding more tagged resources soon, would it make sense to split this into multiple files under a tag related folder?

resin/__init__.py
```diff
@@ -2,8 +2,25 @@
...
+```
+
+From Pip:
+```
+pip install resin-sdk
+```
+or
+```
+pip install git+https://github.com/resin-io/resin-sdk-python.git
+```
```
pimterry: We should make it clear what the recommended approach here is. It's `pip install resin-sdk`, right? We should put that first, and then the others below with some info about why you might want to use this, or drop them entirely if there isn't a clear reason.

resin/logs.py
```diff
@@ -37,7 +37,11 @@ def wrapper(self, *args, **kwargs):
     def subscribe(self, uuid, callback, error):
         """
         This function allows subscribing to device logs.
-        Testing
+        There are fields (`m`, `t`, `s`, `c`) in the output which can be unclear. They stand for:
+            m - The log message itself.
+            t - The log timestamp.
+            s - Is this a system message?
+            c - The id of the service which produced this log (or null).
```
pimterry: It'd be good to explain the null here - `if the device does not support multiple containers`

pimterry (changes requested):
Couple of docs comments, and I'd echo @thgreasi's comments too, but otherwise all the actual code looks good to me.

docs_generator.py
nghiant2710: Updated!

resin/models/tag.py
nghiant2710: Updated! Thanks

resin/models/tag.py
nghiant2710: Updated.

resin/models/tag.py
nghiant2710: If we add more resources here, indeed we should split it but for now i think it will be better to keep it like this for consistency with others (like configvar or envvar)

resin/__init__.py
nghiant2710: Updated!

resin/logs.py
nghiant2710: Updated!

pimterry (approved):

Issue #81, opened 2018-04-27
model: device_environment_variable


Roger:
Is planned to have device_environment_variable resource in models? like in the js sdk?

nghiant2710:
@Roger, that resource is available in Python SDK v3.x and lower. Since Python SDK v4, we support new type of env vars as mentioned in our docs here https://docs.resin.io/learn/manage/serv-vars/#environment-and-service-variables

Roger:
What I don't see, it's support to set environment for all the services in a device in [environment_variables.py](https://github.com/resin-io/resin-sdk-python/blob/master/resin/models/environment_variables.py) (the one in the picture), it's that supported?

![image](https://user-images.githubusercontent.com/9746/39385087-5ad62e02-4a6f-11e8-9150-65c6cb706a08.png)

nghiant2710:
@Roger, they were removed but I've created a PR to re-add them to the models here https://github.com/resin-io/resin-sdk-python/issues/81, will let you know when the PR is merged.

Roger:
thanks! tested that branch and works perfect

nghiant2710:
hey @Roger, just want to let you know that you can upgrade your SDK to v 4.3.0 or higher to have this fix.

PR #84, opened 2018-04-28
Re-add device and application environment variable models

nghiant2710:
We support these type of env vars on the dashboard and we have user asking about these resources so we should bring them back.

Fixes https://github.com/resin-io/resin-sdk-python/issues/81


pimterry (approved):
:+1:

PR #90, opened 2018-05-26
 Implement resin.models.device.get_with_service_details() to get device information with associated services details

### LINKED ISSUE #88: Create function to get details of services running on a device
https://github.com/resin-io/resin-sdk/blob/multicontainer/lib/models/device.coffee#L343

Front logo [Front conversations](https://app.frontapp.com/open/top_blmb)

nghiant2710:
Fix #88


resin/models/device.py
```diff
@@ -184,6 +184,72 @@ def get_by_name(self, name):
...
+            endpoint=self.settings.get('pine_endpoint')
+        )['d']
+
+        if len(raw_data) == 0:
+            raise exceptions.DeviceNotFound(uuid)
+        else:
+            raw_data = raw_data[0]
+
+        raw_data['current_services'] = [self.__get_single_install_summary(i) for i in raw_data['image_install']]
+        raw_data['current_gateway_downloads'] = [self.__get_single_install_summary(i) for i in raw_data['gateway_download']]
```
pimterry: This looks like it should work, but it isn't quite the same as the JS SDK format. Is that intentional?

This code will produce a list of image installs. The SDK output instead produces a map from service names to the image installs for each one. E.g.

```js
current_services: {
	web: [
		{
			id: 1
			service_id: 123
			image_id: 321
			commit: 'new-release-commit'
			status: 'downloading'
			download_progress: 50
		}, {
			id: 2
			service_id: 123
			image_id: 322
			commit: 'old-release-commit'
			status: 'running'
			download_progress: 100
		}
	],
	db: [{
		id: 3
		service_id: 456
		image_id: 323
		commit: 'new-release-commit'
		status: 'running'
		download_progress: 100
	}]
}
```

It's debatable, but personally I think it's a little nicer that way - it makes it more obvious to users that they need to think about multiple instances of the same service and handle them sensibly (because it's a list for each service), and it's a bit more convenient to find the data you're looking for. 

Returning only a single list that users need to filter themselves makes it easier to write code that just finds the first service with the matching service_name and assumes that's the one you want. That'll quickly give you code that works 90% of the time, but implodes if you ever run it while services are updating etc, and it would be nice to help users avoid that.

nghiant2710: @pimterry , I don't have strong opinions about keeping it that way so we should make it consistent with the JS SDK.

pimterry (approved):

PR #92, opened 2018-06-06
Implement functions to set a device or an application to a specific commit

### LINKED ISSUE #91: Implement a feature to set a device or an application to a specific commit
Front logo [Front conversations](https://app.frontapp.com/open/top_c20j)

nghiant2710:
fix #91


resin/models/application.py
```diff
@@ -488,3 +488,34 @@ def generate_provisioning_key(self, app_id):
...
+
+        """
+
+        params = {
+            'filter': 'id',
+            'eq': app_id
+        }
+
+        data = {
+            'commit': commit_id
```
thgreasi, line 516: There is also the `should_track_latest_release` boolean that should be set to `false`, otherwise the `commit` will get updated on the next push.
We also need a way to set this back to `true`.
See: https://github.com/resin-io-playground/staged-releases

thgreasi: Oh I just saw that there is a dedicated method for that.

resin/models/application.py
```diff
@@ -488,3 +488,34 @@ def generate_provisioning_key(self, app_id):
             'POST',
             endpoint=self.settings.get('api_endpoint')
         )
+
+    def set_to_release(self, app_id, commit_id):
+        """
+        Set an application to a specific commit.
```
thgreasi, line 494: How about mentioning that they might also want to call `disable_rolling_updates`?

resin/models/application.py
thgreasi, line 516: Should we error if the commit id is null?

resin/models/device.py
```diff
@@ -1011,3 +1013,37 @@ def revoke_support_access(self, uuid):
...
             endpoint=self.settings.get('pine_endpoint')
         )
+
+    def set_to_release(self, uuid, commit_id):
+        """
+        Set device to a specific release.
+
+        Args:
+            uuid (str): device uuid.
+            commit_id (str) : commit id.
```
thgreasi: This can also be null to restore rolling releases to the device.

thgreasi (commented):
Mostly LGTM :+1:

resin/models/application.py
nghiant2710: 👍

resin/models/application.py
nghiant2710: ah right, I updated the docs to mention this.

resin/models/application.py
nghiant2710: I think we don't need to do it since the API accepts the commit to be null

resin/models/device.py
nghiant2710: 👍 Included in the docs

nghiant2710: And I updated the code to send null release id if the commit id is null so it will work as you said

nghiant2710:
@thgreasi I've fixed your comments, can you please check again?

resin/models/application.py
```diff
@@ -488,3 +488,35 @@ def generate_provisioning_key(self, app_id):
...
+        The commit will get updated on the next push unless rolling updates is disabled (there is a dedicated function for that which is resin.models.applicaion.disable_rolling_updates())
```
thgreasi: NIT: `rolling updates is disabled` => `rolling updates are disabled`
UltraNIT: `dedicated function` => `dedicated method`, since it's a function under a namespace, but no strong opinion on this if it doesn't look Pythony

resin/models/device.py
```diff
@@ -1011,3 +1013,38 @@ def revoke_support_access(self, uuid):
...
         )
+
+    def set_to_release(self, uuid, commit_id):
+        """
+        Set device to a specific release.
+        Set an empty commit_id will restore rolling releases to the device.
+
+        Args:
+            uuid (str): device uuid.
+            commit_id (str) : commit id.
```
thgreasi, line 1024: Should this be `str | None`?

thgreasi (approved):
:+1: but please check my comments wrt the docs before merging

PR #94, opened 2018-06-08
 Allow moving devices to applications of the same arch type

nghiant2710:
Change-type: minor


resin/models/device.py
```diff
@@ -715,7 +715,10 @@ def move(self, uuid, app_name):
         device = self.get(uuid)
         application = self.application.get(app_name)
 
-        if device['device_type'] != application['device_type']:
+        device_dev_type = filter(lambda dev_type: dev_type['slug'] == device['device_type'], self.config.get_device_types())[0]
```
thgreasi: NIT: how about calling `self.config.get_device_types()` once (instead of twice) and keeping the result in a variable?

resin/models/device.py
```diff
@@ -715,7 +715,10 @@ def move(self, uuid, app_name):
...
+        app_dev_type = filter(lambda dev_type: dev_type['slug'] == application['device_type'], self.config.get_device_types())[0]
+
+        if device_dev_type['arch'] != app_dev_type['arch']:
```
thgreasi: NIT: In the JS SDK we also mandate that `device_dev_type.isDependent === app_dev_type.isDependent`

thgreasi (commented):
Mostly LGTM but please address the `isDependent` part for consistency between the SDKs

nghiant2710:
@thgreasi, I can't find the `isDependent` property anywhere that's why I exclude that part, here is the list of device type properties
```
[u'yocto', u'configuration', u'name', u'buildId', u'gettingStartedLink', u'initialization', u'supportsBlink', u'options', u'state', u'version', u'imageDownloadAlerts', u'instructions', u'arch', u'slug', u'isDefault', u'aliases']
```

thgreasi:
@nghiant2710  That's weird since it's definitely in `https://api.resin.io/config` !?!? :alien: :satellite:

thgreasi:
Oh! It might be that the property is only defined for device types that are dependent device types.

nghiant2710:
@thgreasi, updated! 👍

thgreasi (approved):
:+1:

PR #96, opened 2018-07-19
logging: Print deprecation warning for all legacy logs methods (pubnu…

nghiant2710:
…b related) as they will stop working soon

Change-type: patch


resin/logs.py
```diff
@@ -1,11 +1,19 @@
...
 from .base_request import BaseRequest
 from .models.config import Config
 from .models.device import Device
 
 
+logging.basicConfig(format='%(levelname)s:%(message)s')
+
+
+def _print_deprecation_warning():
+    logging.warning(" using legacy logging services, this will stop working shortly.\nPlease update to ensure logs are correctly retrieved in future.")
```
pimterry, line 14: In the JS SDK, we do this only once per process, so if you call a logging method repeatedly you only see the warning the first time.

No biggie, but might be mildly less annoying for users if they're currently using this heavily (or maybe we do want to be annoying, given that their code's going to break). Up to you.

pimterry (approved):
Looks good to me.

We shouldn't release this until just before we merge the new working version though, or users won't have anything to upgrade to.

PR #97, opened 2018-07-20
logging: Add support for new logging endpoint

nghiant2710:


resin/logs.py
```diff
@@ -18,104 +21,50 @@ class Logs(object):
...
+        response = self.base_request.request(
+            '/device/v2/{uuid}/logs'.format(uuid=uuid), 'GET', raw_query=raw_query, stream=True,
+            endpoint=self.settings.get('api_endpoint')
+        )
+
+        if (response.status_code == 200):
+            for line in response.iter_lines(chunk_size=None):
+                if line:
+                    entry = json.loads(line)
+                    callback(entry)
```
pimterry: The stream stays open forever, so this function will never return (unless it throws an error). That's pretty different to the previous behaviour, which returned immediately and called the callback asynchronously. I expect it's also fairly inconvenient from an end user point of view. Can we avoid that?

resin/logs.py
```diff
@@ -18,104 +21,50 @@ class Logs(object):
...
-            ...     print('Error:'+ str(message))
-            Logs.history(uuid=uuid, callback=callback, error=error)
+            count (int): this callback is called on receiving a message from the channel.
 
         """
 
-        channel = self.get_channel(uuid)
-        self.pubnub.history(channel=channel, callback=callback, error=error)
-
-    def unsubscribe(self, uuid):
```
pimterry: This has been removed, and now there's no way for users to unsubscribe at all, right? That seems pretty inconvenient both for end users and us, if it means we end up keeping lots of unwanted connections open.

Looks like it should be fairly easy to implement something to do this with just `response.close()`. An API design that allows that is a trickier question though... In the JS SDK, we return an object from subscribe, with `on('line', (line) => ...)` methods to listen to events, plus an `unsubscribe()` method to kill the subscription & connection entirely.

I'm not really up to date on idiomatic Python, but I imagine something similar might fit well here.

resin/logs.py
```diff
@@ -18,104 +21,50 @@ class Logs(object):
...
-                self.pubnub = Pubnub(
-                    publish_key=pubnub_key['publish_key'],
-                    subscribe_key=pubnub_key['subscribe_key']
-                )
-            return func(self, *args, **kwargs)
-        return wrapper
-
-    @_init_pubnub
-    def subscribe(self, uuid, callback, error):
+    def subscribe(self, uuid, callback):
```
pimterry: There's actually a `count` option here as well.

pimterry (dismissed):

resin/logs.py
nghiant2710: @pimterry, this is the same behaviour as the pubnub SDK before. Should I keep the old behaviour or make it consistent with the JS SDK?

resin/logs.py
nghiant2710: Will add this asap.

resin/logs.py
nghiant2710: @pimterry, I don't see a count option here https://github.com/resin-io/resin-sdk/blob/master/lib/logs.coffee#L164 or is it added in another branch?

pimterry: Yep, it isn't merged in the SDK yet, I only got reminded about it last night. SDK PR is here: https://github.com/resin-io/resin-sdk/pull/548

resin/logs.py
nghiant2710: The Python V4 Pubnub is asynchronous so will update this

pimterry: > this is the same behaviour as the pubnub SDK before

That's not true. With the currently published SDK, you can run:

```python
>>> resin.logs.subscribe('30579e3822891248ae54fdd4f626129a', callback, error)
>>> 
```

The function has returned there after starting the subscription, and then calls callback/error asynchronously later.

This code never does, and will stay waiting in this loop and blocking this thread forever, never returning.

pimterry: Sorry, forgot the conclusion: yes, I think we should make this async, matching the previous behaviour & the JS SDK.

resin/base_request.py
```diff
@@ -42,7 +42,9 @@ def __set_authorization(self, headers):
             {'Authorization': 'Bearer {:s}'.format(self.settings.get(TOKEN_KEY))})
 
     def __get(self, url, headers, data=None, stream=None):
-        return requests.get(url, headers=headers, timeout=self.timeout)
+        if not stream:
+            return requests.get(url, headers=headers, timeout=self.timeout)
+        return requests.get(url, headers=headers, stream=stream)
```
petrosagg: Why don't you always pass stream to `requests`? No need for the if statement

setup.py
```diff
@@ -73,7 +73,7 @@ def get_version(package):
     # your project is installed. For an analysis of "install_requires" vs pip's
     # requirements files see:
     # https://packaging.python.org/en/latest/requirements.html
-    install_requires=['PyJWT', 'requests', 'pubnub<4.0.0', 'pyotp'],
+    install_requires=['PyJWT>=1.5.0', 'requests>=2.19.1', 'pyotp>=2.2.5', 'pyOpenSSL==18.0.0', 'Twisted>=18.7.0', 'service-identity==17.0.0', 'google-auth-oauthlib>=0.2.0'],
```
petrosagg: what do we need `google-auth-oauthlib` and `service-identity` for?

resin/logs.py
```diff
@@ -1,133 +1,189 @@
...
+
+class Subscription(object):
     """
-    This class implements functions that allow processing logs from device.
+    This is low level class and is not meant to be used by end users directly.
+    """
+
+    def __init__(self):
+        self.context_factory = WebClientContextFactory()
+        self.settings = Settings()
```
petrosagg, line 85: unrelated to this specific PR but needs to be addressed. Settings should be a singleton object. Now it's instantiated many many times causing the config file to be re-parsed and re-written every single time. Put a `print` in the module to see what I mean

resin/logs.py
```diff
@@ -1,133 +1,189 @@
...
+        super(Subscription, self).__init__()
+
+    def add(self, uuid, callback, error=None, count=None):
+
```
petrosagg: NIT: extra new line here

resin/logs.py
```diff
@@ -1,133 +1,189 @@
...
+        if count:
+            query = 'stream=1&count={}'.format(count)
 
-    This class is implemented using pubnub python sdk.
+        url = urljoin(
+            self.settings.get('api_endpoint'),
+            '/device/v2/{uuid}/logs?{query}'.format(uuid=uuid, query=query)
+        )
+        headers = {}
+        headers[b'Authorization'] = ['Bearer {:s}'.format(self.settings.get('token')).encode()]
```
petrosagg, line 97: Why is the value of the header a list?

petrosagg (changes requested):

resin/base_request.py
nghiant2710: yeah, that's right, updated!

setup.py
nghiant2710: pyOpenSSL, service-identity, google-auth-oauthlib are dependencies for twisted, otherwise, we will see some ugly SSL certificate warnings when sending https request.

resin/logs.py
nghiant2710: ack, I already opened an issue for this before but haven't fixed yet, will fix it soon.

resin/logs.py
nghiant2710: it's how twisted defines a header, `_rawHeaders | A dict mapping header names as str to lists of header values as str.` (https://twistedmatrix.com/documents/9.0.0/api/twisted.web.http_headers.Headers.html)


`

setup.py
petrosagg: surely `service-identity` and `google-auth-oauthlib` are optional dependencies and can be removed.

resin/logs.py
```diff
@@ -1,133 +1,188 @@
...
-        Examples:
-            # Define callback and error.
-            >>> def callback(message):
-            ...     print(message)
-            >>> def error(message):
-            ...     print('Error:'+ str(message))
-            Logs.history(uuid=uuid, callback=callback, error=error)
 
-        """
+        self.device.get(uuid)
```
petrosagg: What is this call doing? I don't see it used anywhere. Also here you seem to be using `uuid` but right afterwards checking if it's None

resin/logs.py
```diff
@@ -1,133 +1,188 @@
...
+class Subscription(object):
     """
-    This class implements functions that allow processing logs from device.
+    This is low level class and is not meant to be used by end users directly.
+    """
+
+    def __init__(self):
+        self.context_factory = WebClientContextFactory()
+        self.settings = Settings()
+        super(Subscription, self).__init__()
```
petrosagg: Is there any reason for this since you inherit from object? Why do we need an explicit inherit from object?

resin/logs.py
```diff
@@ -1,133 +1,188 @@
...
-            # Define callback and error.
-            >>> def callback(message):
-            ...     print(message)
-            >>> def error(message):
-            ...     print('Error:'+ str(message))
-            Logs.history(uuid=uuid, callback=callback, error=error)
 
-        """
+        self.device.get(uuid)
+        if uuid is not None and uuid not in self.subscriptions:
```
petrosagg: This will not work if a library user calls subscribe twice with two different callbacks

resin/logs.py
```diff
@@ -1,133 +1,188 @@
...
+        if count:
+            raw_query = 'count={}'.format(count)
 
-    def get_channel(self, uuid):
+        return self.base_request.request(
+            '/device/v2/{uuid}/logs'.format(uuid=uuid), 'GET', raw_query=raw_query,
+            endpoint=self.settings.get('api_endpoint')
+        )
+
+    def unsubscribe(self, uuid=None):
```
petrosagg: uuid should be mandatory when unsubscribing. if you want to add this functionality I suggest adding an `unsubscribe_all` method

resin/logs.py
```diff
@@ -1,133 +1,188 @@
...
-            device_info = self.device.get(uuid)
-            if 'logs_channel' in device_info:
-                self.logs_channel = device_info['logs_channel']
-            else:
-                self.logs_channel = uuid
-
-        return 'device-{logs_channel}-logs'.format(logs_channel=self.logs_channel)
+        if uuid and uuid in self.subscriptions:
+            self.subscription_handler.stop(self.subscriptions[uuid])
+            self.subscriptions.pop(uuid, None)
```
petrosagg: No need to `pop` if you're not using the result. Use `del self.subscriptions[uuid]` instead

resin/logs.py
```diff
@@ -1,133 +1,188 @@
...
 
-        """
+        self.device.get(uuid)
+        if uuid is not None and uuid not in self.subscriptions:
+            self.subscriptions[uuid] = self.subscription_handler.add(uuid, callback, error, count)
 
-        _print_deprecation_warning()
-        channel = self.get_channel(uuid)
-        self.pubnub.history(channel=channel, callback=callback, error=error)
+        self.subscription_handler.run()
```
petrosagg: This implementation still spawns one thread per connection. That's not how twisted should be used and ideally you shouldn't be dealing with threads at all.

petrosagg (changes requested):

setup.py
nghiant2710: yeah, moved them to optional dependencies section.

resin/logs.py
nghiant2710: this is uuid validation, this call gets device info from the API and DeviceNotFound exception will be raised if the uuid is invalid or user doesn't have access to a device. 

Removed the check for None since it's no longer necessary.

resin/logs.py
nghiant2710: oh, this is just a leftover since this class inherited from Thread before we change to twisted. Removed!

resin/logs.py
nghiant2710: Fixed, users now can subscribe for device logs with different callbacks

resin/logs.py
nghiant2710: yeah, it makes sense to have an `unsubscribe_all` method.

resin/logs.py
nghiant2710: IMO, I prefer this way since it won't raise any exception if uuid doesn't exist in the dict but we already have a check for that case so it's fine.
Updated.

resin/logs.py
nghiant2710: This call won't spawn a new thread per connection, it only kicks off the reactor if it's not running.

resin/logs.py
```diff
@@ -1,133 +1,194 @@
...
             uuid (str): device uuid.
 
         """
 
-        _print_deprecation_warning()
-        if hasattr(self, 'pubnub'):
-            channel = self.get_channel(uuid)
-            self.pubnub.unsubscribe(channel=channel)
+        if uuid and uuid in self.subscriptions:
+            [self.subscription_handler.stop(d) for d in self.subscriptions[uuid]]
```
petrosagg: Since you're using the comprehension there is no reason to not do a normal for loop

resin/logs.py
```diff
@@ -1,133 +1,194 @@
...
+        Returns:
+            dict: a log entry will contain the following keys: `isStdErr, timestamp, message, isSystem, createdAt`.
 
         """
 
-        _print_deprecation_warning()
-        channel = self.get_channel(uuid)
-        self.pubnub.subscribe(channels=channel, callback=callback, error=error)
+        self.device.get(uuid)
+        if uuid not in self.subscriptions:
```
petrosagg: a better pattern for this is to separate initialisation from the general case like so:

```python
if uuid not in self.subscriptions:
    self.subscriptions[uuid] = []

self.subscriptions[uuid].append(self.subscription_handler.add(uuid, callback, error, count))
```

or you can make `self.subscriptions` be a `defaultdict(list)` and skip the if completely. This is probably the most pythonic way of doing it.

https://docs.python.org/2/library/collections.html#collections.defaultdict

resin/logs.py
```diff
@@ -1,133 +1,194 @@
...
+    """
 
-# TODO: https://github.com/resin-io/resin-sdk/pull/277/files
-class Logs(object):
+    def __init__(self, callback, error):
+        self.callback = callback
+        self.error = error
+
+    def dataReceived(self, data):
+        d = Deferred()
```
petrosagg: Why do you construct a Deferred here? You can process the data immediately

resin/logs.py
```diff
@@ -1,133 +1,194 @@
...
+        d.addCallback(self.display_data)
+        d.addErrback(self.error_func)
+        d.callback(data)
+
+    def connectionLost(self, reason):
```
petrosagg: You should call the user's error callback in this case

resin/logs.py
```diff
@@ -1,133 +1,194 @@
...
+        pass
+
+    def display_data(self, data):
+        lines = data.splitlines()
```
petrosagg: this will drop logs if `data` contains a half finished line like this

```python
data = '{"message": "foobar"}\n{"mess'

File: resin/logs.py [hunk 20]
```diff
@@ -1,133 +1,194 @@
...
+        if self.error:
+            d = Deferred()
+            d.addErrback(self.error)
+            d.errback(DefaultException(error))
+
+
+def cbRequest(response, callback, error):
+    protocol = StreamingParser(callback, error)
+    response.deliverBody(protocol)
+    return protocol
```
petrosagg, line 70: no need to return here

resin/logs.py
```diff
@@ -1,133 +1,194 @@
...
+                self.callback(json.loads(line))
+            except:
+                # Ignore incomplete log entry, this can happen with the old logs
+                pass
+
+    def error_func(self, error):
+        if self.error:
+            d = Deferred()
+            d.addErrback(self.error)
+            d.errback(DefaultException(error))
```
petrosagg: no need for Deferred here either. You can call self.error directly with the exception

resin/logs.py
```diff
@@ -1,133 +1,194 @@
...
+        d = agent.request(b'GET', url.encode(), Headers(headers), None)
+        d.addCallback(cbRequest, callback, error)
+        return d
+
+    def run(self):
+        if not reactor.running:
+            Thread(target=reactor.run, args=(False,)).start()
+
+    def stop(self, d):
+        d.addCallback(cbDrop)
```
petrosagg: Will this be called from the reactor thread? It's unsafe if it's not

resin/logs.py
```diff
@@ -1,133 +1,194 @@
...
-        if not hasattr(self, 'logs_channel'):
-            device_info = self.device.get(uuid)
-            if 'logs_channel' in device_info:
-                self.logs_channel = device_info['logs_channel']
-            else:
-                self.logs_channel = uuid
-
-        return 'device-{logs_channel}-logs'.format(logs_channel=self.logs_channel)
+        for device in self.subscriptions:
+            [self.subscription_handler.stop(d) for d in self.subscriptions[device]]
```
petrosagg: Same here, better to use a normal loop

setup.py
```diff
@@ -82,6 +82,7 @@ def get_version(package):
     extras_require={
         'dev': ['check-manifest'],
         'test': ['coverage'],
+        'ssl': ['service-identity']
```
petrosagg: Why do you require `service-identity`?

petrosagg (changes requested):

resin/logs.py
nghiant2710: Updated.

resin/logs.py
nghiant2710: `defaultdict` will be the best choice here, updated.

resin/logs.py
nghiant2710: updated

setup.py
nghiant2710: this is optional dependency not a required one (it will be installed if user specifies extras .e.g, `pip install -e .[ssl]`)

I still keep it in the `setup.py` since Python 3.x will print a warning like this `:0: UserWarning: You do not have a working installation of the service_identity module: 'No module named 'service_identity''.  Please install it from <https://pypi.python.org/pypi/service_identity> and make sure all of its dependencies are satisfied.  Without the service_identity module, Twisted can perform only rudimentary TLS client hostname verification.  Many valid certificate/hostname mappings may be rejected.` if `service-identity` is not installed.

resin/logs.py
nghiant2710: The connectionLost is called when the body has been completely delivered or the connection is shut down (after `loseConnection()` is called and we call this method on unsubscribing) and any tearing down of Protocol-specific objects is done here so nothing relates to errors should be passed to user's error callback in this case.

resin/logs.py
nghiant2710: If in the code, we don't spawn a new thread by doing `reactor.callinThread()` then no need to worry about this. Every call to reactor API will be run in the reactor thread.

resin/logs.py
nghiant2710: Yeah, I tidied up the code here and it's easier to read now.

resin/logs.py
nghiant2710: Since the protocol controls the flow of the response data after `response.deliveryBody` is called, the protocol should be passed to unsubscribe callback so that we can shut down the connection on unsubscribing.

resin/logs.py
nghiant2710: updated the behavior here, so user's error callback will be called in case of data loss or half finished data received (in this case, `{"message": "foobar"}` will be passed to user's callback and Failure contains ValueError exception will be passed to the error callback )

However, I don't have a strong opinion  here since I'm not so sure about the chunk sent by the API so if you think it's better to have a pending that stores unfinished log which will be merged with next chunk then I will update this.

setup.py
petrosagg: in this case `service-identity` should be mandatory

resin/logs.py
petrosagg: stop is not called by the reactor thread. it's called by the main thread

resin/logs.py
petrosagg: of course we should keep the pending data. Why would any user care that the line was received in two chunks?

resin/logs.py
petrosagg: You're still using a Deferred for no reason. Just call the error callback directly

resin/logs.py
nghiant2710: Changed to a thread-safe method.

setup.py
nghiant2710: Moved it to required dependencies.

requirements.txt
```diff
@@ -0,0 +1,5 @@
+PyJWT>=1.5.0
+requests>=2.19.1
```
petrosagg, line 2: missing service-identity here

resin/logs.py
```diff
@@ -1,133 +1,190 @@
...
+        self.pending = ''
+
+    def dataReceived(self, data):
+        self.pending += data
+        lines = self.pending.splitlines()
+        try:
+            json.loads(lines[-1])
+            self.pending = ''
+        except ValueError:
+            self.pending = lines[-1]
```
petrosagg: This is not correct. The code will not differentiate between a pending line and garbage coming it, leading to accumulating garbage instead of calling the error callback.

Here is a correct way to do it:

```python
self.pending += data

lines = self.pending.split('\n')
self.pending = lines.pop()

for line in lines:
    try:
        obj = json.loads(line)
    except Exception as e:
        if self.error:
            self.error(e)
        # TODO: close connection here and stop processing
        break
        
    self.callback(obj)
```

resin/logs.py
```diff
@@ -1,133 +1,190 @@
...
+            dict: a log entry will contain the following keys: `isStdErr, timestamp, message, isSystem, createdAt`.
 
         """
 
-        _print_deprecation_warning()
-        channel = self.get_channel(uuid)
-        self.pubnub.subscribe(channels=channel, callback=callback, error=error)
+        self.device.get(uuid)
+        self.subscriptions[uuid].append(self.subscription_handler.add(uuid, callback, error, count))
+        self.subscription_handler.run()
```
petrosagg: This looks like an implementation detail of the Subscription class. Better call it within the `add` method

petrosagg (changes requested):

resin/logs.py
nghiant2710: updated!

requirements.txt
nghiant2710: updated!

resin/logs.py
nghiant2710: updated.

resin/logs.py
```diff
@@ -1,133 +1,191 @@
...
+        self.pending += data
+
+        lines = self.pending.split(b'\n')
+        self.pending = lines.pop()
+
+        for line in lines:
+            try:
+                if line:
+                    obj = json.loads(line)
+                    self.callback(obj)
```
petrosagg: you shouldn't call the user callback here. if it raises an exception it's the user's responsibility to handle it, not of this piece of code

resin/logs.py
```diff
@@ -1,133 +1,191 @@
...
+            except Exception as e:
+                if self.error:
+                    self.error(e)
+                    self.transport.stopProducing()
```
petrosagg: You should stop producing and close the connection regardless of whether an error callback is registered

petrosagg (changes requested):

petrosagg (approved):

PR #116, opened 2018-10-15
Allow overriding the test API endpoint

### LINKED ISSUE #115: Allow overriding the API endpoint in tests
This will make it much easier to test the renamed SDK before the new API has reached prod.

pimterry:
This is going to make it easier to start early testing against the new API, once that's available.

I've also added some docs, and fixed the app name test to work against the code currently on staging (it currently fails against prod due to a prod bug, but the fix that's on staging slightly changes the actual error message).

Fixes #115

### DIFF
--- a/tests/helper.py
+++ b/tests/helper.py
@@ -25,6 +25,11 @@ class TestHelper(object):
         self.resin_exceptions = resin_exceptions
         self.base_request = BaseRequest()
         self.settings = Settings()
+
+        if 'api_endpoint' in self.credentials and self.credentials['api_endpoint'] is not None:
+            self.settings.set('api_endpoint', self.credentials['api_endpoint'])
+            self.settings.set('pine_endpoint', '{0}/{1}/'.format(self.credentials['api_endpoint'], self.settings.get('api_version')))
+
         if not self.resin.auth.is_logged_in():
             self.resin.auth.login(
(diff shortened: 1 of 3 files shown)


tests/helper.py
```diff
@@ -25,6 +25,11 @@ def __init__(self):
         self.resin_exceptions = resin_exceptions
         self.base_request = BaseRequest()
         self.settings = Settings()
+
+        if 'api_endpoint' in self.credentials:
+            self.settings.set('api_endpoint', self.credentials['api_endpoint'])
+            self.settings.set('pine_endpoint', '%s/%s/' % (self.credentials['api_endpoint'], self.settings.get('api_version')))
```
nghiant2710: You can use '`{0}/{1}'.format(self.credentials['api_endpoint'], self.settings.get('api_version'))` here instead to ensure the consistency with other components.

nghiant2710 (approved):

nghiant2710:
@pimterry, LGTM 👍. Can you please link the API PR with this one so I can merge it when the new API is available?

pimterry:
@nghiant2710 the API PR is this one: https://github.com/resin-io/resin-api/pull/1541

pimterry:
API has actually now been released, we're good to go!

PR #117, opened 2018-10-16
Rename the Python SDK from 'resin' to 'balena'

### LINKED ISSUE #113: Rename the Python SDK

pimterry:
Fixes #113 

This is a breaking change in lots of places, but nothing structural, it's just renaming.

Where possible, I've tried to keep compatibility for existing RESIN_ configurations, e.g. when running on a device, we now read the supervisor api key/address from whichever is available of `BALENA_SUPERVISOR_API_KEY` and `RESIN_SUPERVISOR_API_KEY`, and we accept both `BALENA_API_KEY` and `RESIN_API_KEY` elsewhere.

There is one failing test here, as the API currently doesn't reject `io.balena.*` tags. That should get resolved once the new API is available (which is a prerequisite for releasing this anyway).

I've also had to fix the docs generation slightly - it was failing because BaseTag defines a `get` method without any docstring. As far as I can tell that's not supposed to be used. It doesn't take UUIDs like the other methods for example, and of course it's not included in the docs, so I've just made it an internal `_get` instead.


nghiant2710 (approved):

nghiant2710:
Great work @pimterry 👍 Sorry, I should have done this task but I'm current busy with the base images. Thanks a lot

pimterry:
No worries! It fits in very neatly with the JS SDK work :+1:

We should talk about how publishing works for the python SDK at some point though, come grab me tomorrow some time and we can we chat about it.

PR #121, opened 2018-11-20
Balena Python SDK

nghiant2710:
Connects to https://github.com/balena-io/balena-sdk/pull/597, https://github.com/balena-io/balena-sdk/pull/572.

This PR contains some breaking changes:
- Update to API v5.
- Using some Opensource endpoints.
- Add some release checking methods for staged release support.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -212,6 +212,26 @@ Get a single application by application id.
 >>> balena.models.application.get_by_id(9020)
 {u'app_name': u'RPI1', u'__metadata': {u'type': u'', u'uri': u'/ewa/application(9020)'}, u'git_repository': u'g_trong_nghia_nguyen@git.balena.io:g_trong_nghia_nguyen/rpi1.git', u'user': {u'__deferred': {u'uri': u'/ewa/user(5397)'}, u'__id': 5397}, u'device_type': u'raspberry-pi', u'commit': None, u'id': 9020}
 ```
+### Function: get_by_owner(name, owner)
+
+Get a single application.
+
+#### Args:
+    name (str): application name.
+    owner (str):  owner's username.
+
+#### Returns:
+    dict: application info.
+
+#### Raises:
+    ApplicationNotFound: if application couldn't be found.
+    AmbiguousApplication: when more than one application is returned.
+
+#### Examples:
+```python
+>>> balena.models.application.get_by_owner('mothaiba', 'pythonsdk_test_resin')
+{u'depends_on__application': None, u'should_track_latest_release': True, u'app_name': u'mothaiba', u'application_type': {u'__deferred': {u'uri': u'/resin/application_type(5)'}, u'__id': 5}, u'__metadata': {u'type': u'', u'uri': u'/resin/application(1307755)'}, u'is_accessible_by_support_until__date': None, u'actor': 3438708, u'id': 1307755, u'user': {u'__deferred': {u'uri': u'/resin/user(32986)'}, u'__id': 32986}, u'device_type': u'raspberrypi3', u'commit': None, u'slug': u'pythonsdk_test_resin/mothaiba'}
+```
 ### Function: get_config(app_id)
 
         Download application config.json.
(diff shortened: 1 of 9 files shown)


balena/models/application.py
```diff
@@ -517,10 +554,108 @@ def set_to_release(self, app_id, commit_id):
...
+
+        Examples:
+            >> > balena.models.application.will_track_new_releases('5685')
+            True
+
+        """
+
+        return bool(self.get_by_id(app_id)['should_track_latest_release'])
+
+    def is_track_latest_release(self, app_id):
```
dfunckt: I think better names for this one and the above are:

- is_track_latest_release  -> is_on_latest_release
- will_track_new_releases  -> is_tracking_latest_release

balena/models/release.py
```diff
@@ -81,6 +106,24 @@ def get_all_by_application(self, app_id):
...
+
+        Args:
+            app_id (str): applicaiton id.
+
+        Returns:
+            dict: release info.
+
+        """
+
+        raw_query = "$top=1&$filter=belongs_to__application%20eq%20'{}'%20and%20status%20eq%20'success'".format(app_id)
```
dfunckt, line 121: What is `belongs_to__application eq '{}'`?

balena/models/application.py
nghiant2710: @dfunckt i have the same thought but I think it's better to respect and use the same name for same methods with the JS SDK.

dfunckt: @nghiant2710 where are these defined on JS SDK? Github search doesn't bring up any result.

balena/models/release.py
nghiant2710: it's a formatted string resolves to `belongs_to__application eq 'app_id'`. I've not had time to implement something the same as pineJS in Python so we're using raw OData query for some complex query in the Python SDK.

balena/models/application.py
dfunckt: Oh, NVM, I search for the exact strings, but on JS they're camelCase.

nghiant2710: https://github.com/balena-io/balena-sdk/blob/master/lib/models/application.coffee#L683

dfunckt: @nghiant2710 `is_track_latest_release` isn't correct English, and in fact [in JS SDJ it's `isTrackingLatestRelease`](https://github.com/balena-io/balena-sdk/blob/master/lib/models/application.coffee#L739). Even though I think the method names are bad, I agree with you that we should use the same names, but let's fix `is_track_latest_release`.

balena/models/release.py
dfunckt: Yeah, I forgot about Python string formatting, sorry about that, ignore my comment.

balena/models/application.py
nghiant2710: @dfunckt ah yes, my typo, the same method in device class is correct but this one is wrong. Fixed 👍

DOCUMENTATION.md
```diff
@@ -212,6 +212,26 @@ Get a single application by application id.
...
+    dict: application info.
+
+#### Raises:
+    ApplicationNotFound: if application couldn't be found.
+    AmbiguousApplication: when more than one application is returned.
+
+#### Examples:
+```python
+>>> balena.models.application.get('RPI1')
+{u'app_name': u'RPI1', u'__metadata': {u'type': u'', u'uri': u'/ewa/application(9020)'}, u'git_repository': u'g_trong_nghia_nguyen@git.balena.io:g_trong_nghia_nguyen/rpi1.git', u'user': {u'__deferred': {u'uri': u'/ewa/user(5397)'}, u'__id': 5397}, u'device_type': u'raspberry-pi', u'commit': None, u'id': 9020}
```
shaunmulligan: @nghiant2710 what does this example have `/ewa/application(9020)` in it?

nghiant2710: @shaunmulligan fixed

balena/models/config.py
```diff
@@ -50,9 +50,11 @@ def get_all(self):
 
         """
 
-        if self._config is None:
+        if not self._config:
             self._config = self.base_request.request(
                 'config', 'GET', endpoint=self.settings.get('api_endpoint'))
```
dfunckt, line 56: This request will get a 404 from an open source API because there's no `/config` endpoint, and it seems to me you're not handling it.

nghiant2710: @dfunckt the 404 is handled here https://github.com/balena-io/balena-sdk-python/blob/master/balena/base_request.py#L190-L191.
However, it still breaks since `get_all()` will always throws an exception if open source API is used and users won't be able to fetch the device types. To fix this, I made a small change and `get_all()` will only fetch the configuration not the device types.

dfunckt (approved):

PR #123, opened 2018-12-08
application: Add options parameter to application.get_config()

### LINKED ISSUE #122: Application.get_config errors out with HTTP 400 - 'A version is required.'
We need to pass an additional request parameter for OS version to the API - `version`.

Front logo [Front conversations](https://app.frontapp.com/open/top_fwqr)

nghiant2710:
the version option is required in options parameter.
Fixes #122 

Change-type: minor


majorz:
I think we need to move out the `version` parameter out of the `options` dict and make it the second argument, since it is mandatory and not optional.

Also it will feel more pythonic to have the `options` argument be a keywords argument - `**options`. And have the members like `wifiKey` converted to a standard notation - `wifi_key`. Not sure how that would align with the overall API though in terms of consistency.

nghiant2710:
@majorz, I agree with making `version` a mandatory argument and options as a keyword argument. However, The available options of `options` args shouldn't be changed since I think the API only accepts the camel case naming and it should be aligned with the JS SDK

majorz:
@nghiant2710 we can solve this with a utility function that iterates through all keys in `options` and converts their notation from Python to JS. This will add only one additional line in the function body or it can be done as a decorator. My thinking is that the the Python API does not necessary have to match precisely the internal remote JS API in the naming convention. One is the public API, the other one is internal implementation detail in this case. But if we already use the naming convention like that in other places, we better leave it like that for consistency.

nghiant2710:
@majorz yup, we use that naming convention in other places so I think it's better to keep using it for this method. I will note this down and maybe have another PR to convert all naming convention in a more pythonic way.

nghiant2710:
ping @majorz, can you approve this PR please?

nghiant2710:
@shaunmulligan, @majorz can I have a review please?

majorz (approved):
Sorry for the delay!

PR #132, opened 2019-04-02
device: Add get_supervisor_state device method

nghiant2710:
Connects to: https://github.com/balena-io/balena-sdk/pull/648
HQ: balena-io/balena#1663

Change-type: minor

### DIFF
--- a/tests/functional/models/test_device.py
+++ b/tests/functional/models/test_device.py
@@ -362,6 +362,18 @@ class TestDevice(unittest.TestCase):
         self.balena.models.device.disable_lock_override(device['uuid'])
         self.assertFalse(self.balena.models.device.has_lock_override(device['uuid']))
 
+    def test_get_supervisor_state(self):
+        device = self.helper.create_device()[1]
+
+        # should be rejected if the device doesn't exist.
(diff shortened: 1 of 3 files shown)


xginn8 (approved):
lgtm!

tests/functional/models/test_device.py
```diff
@@ -362,6 +362,18 @@ def test_disable_lock_override(self):
         self.balena.models.device.disable_lock_override(device['uuid'])
         self.assertFalse(self.balena.models.device.has_lock_override(device['uuid']))
 
+    def test_get_supervisor_state(self):
+        app, device = self.helper.create_device()
```
xginn8: why create the `app` variable if it isn't used? you can just drop it with `_` if `create_device()` returns a tuple

nghiant2710: @xginn8 ah yeah, I copied it from another test so didn't notice about that.

I think it's better to use `device = self.helper.create_device()[1]` here if we want to drop the unused `app` variable since `_` has few conventional uses, not just a "throwaway" variable so probably should avoid using it if not necessary.

I updated the code, btw

imrehg (approved):
LGTM

PR #133, opened 2019-04-05
hup: Add device OS update methods

nghiant2710:
Connects to https://github.com/balena-io/balena-sdk/pull/638
HQ: balena-io/balena#1215

Change-type: minor

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -912,6 +912,23 @@ Get device name by device uuid.
 
 #### Raises:
     DeviceNotFound: if device couldn't be found.
+### Function: get_os_update_status(uuid)
+
+        Get the OS update status of a device.
+
+####         Args:
+            uuid (str): device uuid.
+
+####         Returns:
+            dict: action response.
+
+####         Examples:
+```python
+            >>> balena.models.device.get_os_update_status('b6070f4fea5edf808b576123157fe5ec')
+            {u'status': u'done', u'parameters': {u'target_version': u'2.29.2+rev1.prod'}, u'stdout': u'[1554490814][LOG]Normalized target version: 2.29.2+rev1
+', u'last_run': 1554491107242L, u'error': u'', u'action': u'resinhup'}
+        
+```
 ### Function: get_status(uuid)
 
 Get the status of a device.
@@ -958,7 +975,7 @@ Get the supervisor target state on a device
 #### Examples:
 ```python
 >>> balena.models.device.get_supervisor_target_state('b6070f4fea5edf808b576123157fe5ec')
-{u'local': {u'name': u'holy-darkness', u'config': {u'RESIN_SUPERVISOR_NATIVE_LOGGER': u'true', u'RESIN_SUPERVISOR_POLL_INTERVAL': u'900000'}, u'apps': {u'1398898': {u'name': u'test-nuc', u'commit': u'f9d139b80a7df94f90d7b9098b1353b14ca31b85', 6 u'releaseId': 850293, u'services': {u'229592': {u'imageId': 1016025, u'serviceName': u'main', u'image': u'registry2.balena-cloud.com/v2/27aa30131b770a4f993da9a54eca6ed8@sha256:f489c30335a0036ecf1606df3150907b32ea39d73ec6de825a549385022e3e22', u'running': True, u'environment': {}, u'labels': {u'io.resin.features.dbus': u'1', u'io.resin.features.firmware': u'1', u'io.resin.features.kernel-modules': u'1', u'io.resin.features.resin-api': u'1', u'io.resin.features.supervisor-api': u'1'}, u'privileged': True, u'tty': True, u'restart': u'always', u'network_mode': u'host', u'volumes': ['resin-data:/data']}}, 6 u'volumes': {u'resin-data': {}}, u'networks': {}}}}, u'dependent': {u'apps': {}, u'devices': {}}}
+{u'local': {u'name': u'holy-darkness', u'config': {u'RESIN_SUPERVISOR_NATIVE_LOGGER': u'true', u'RESIN_SUPERVISOR_POLL_INTERVAL': u'900000'}, u'apps': {u'1398898': {u'name': u'test-nuc', u'commit': u'f9d139b80a7df94f90d7b9098b1353b14ca31b85', u'releaseId': 850293, u'services': {u'229592': {u'imageId': 1016025, u'serviceName': u'main', u'image': u'registry2.balena-cloud.com/v2/27aa30131b770a4f993da9a54eca6ed8@sha256:f489c30335a0036ecf1606df3150907b32ea39d73ec6de825a549385022e3e22', u'running': True, u'environment': {}, u'labels': {u'io.resin.features.dbus': u'1', u'io.resin.features.firmware': u'1', u'io.resin.features.kernel-modules': u'1', u'io.resin.features.resin-api': u'1', u'io.resin.features.supervisor-api': u'1'}, u'privileged': True, u'tty': True, u'restart': u'always', u'network_mode': u'host', u'volumes': ['resin-data:/data']}}, u'volumes': {u'resin-data': {}}, u'networks': {}}}}, u'dependent': {u'apps': {}, u'devices': {}}}
 ```
 ### Function: get_supported_device_types()
 
@@ -1214,18 +1231,44 @@ Set an empty commit_id will restore rolling releases to the device.
 ### Function: set_to_release_by_id(uuid, release_id)
 
 Set device to a specific release by release id (please notice that release id is not the commit hash on balena dashboard).
-Set an empty commit_id will restore rolling releases to the device.
+Remove release_id will restore rolling releases to the device.
 
 #### Args:
     uuid (str): device uuid.
-    release_id (str) : build id.
+    release_id (Optional[int]): release id.
 
 #### Returns:
     OK.
 
 #### Examples:
-    >> > balena.models.device.set_to_release_by_id('49b2a76b7f188c1d6f781e67c8f34adb4a7bfd2eec3f91d40b1efb75fe413d', '165432')
-    'OK'
+```python
+>>> balena.models.device.set_to_release_by_id('49b2a76b7f188c1d6f781e67c8f34adb4a7bfd2eec3f91d40b1efb75fe413d', 165432)
+'OK'
+>>> balena.models.device.set_to_release_by_id('49b2a76b7f188c1d6f781e67c8f34adb4a7bfd2eec3f91d40b1efb75fe413d')
+'OK'
+```
+### Function: start_os_update(uuid, target_os_version)
+
+Start an OS update on a device.
+
+#### Args:
+    uuid (str): device uuid.
+    target_os_version (str): semver-compatible version for the target device.
+        Unsupported (unpublished) version will result in rejection.
+        The version **must** be the exact version number, a "prod" variant and greater than the one running on the device.
+
+#### Returns:
+    dict: action response.
+
+#### Raises:
+    DeviceNotFound: if device couldn't be found.
+    InvalidParameter|OsUpdateError: if target_os_version is invalid.
+
+#### Examples:
+```python
+>>> balena.models.device.start_os_update('b6070f4fea5edf808b576123157fe5ec', '2.29.2+rev1.prod')
+{u'status': u'in_progress', u'action': u'resinhup', u'parameters': {u'target_version': u'2.29.2+rev1.prod'}, u'last_run': 1554490809219L}
+```
 ### Function: track_application_release(uuid)
 
 Configure a specific device to track the current application release.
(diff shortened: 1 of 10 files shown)


DOCUMENTATION.md
```diff
@@ -1251,6 +1290,22 @@ Get an application config.json.
 
 #### Returns:
     dict: application config.json
+### Function: get_device_os_semver_with_variant(os_version, os_variant)
```
imrehg, line 1344: @nghiant2710 can we have an example for this function call as well?

nghiant2710: @imrehg, done, can you please check this PR again?

imrehg: That part is good, thanks!

balena/models/hup.py
```diff
@@ -0,0 +1,238 @@
+import semver
+
+from .. import exceptions
+
+
+class Hup(object):
```
imrehg, line 6: @nghiant2710 why is this part really needed? I would rather not hardcode the device types here as well, as we are trying to remove in other tools too. The proxy already gatekeeping the allowed versions, and in the NodeSDK it's just a stopgap measure to remove these hardcoded values. 

In generally also IMHO do not concern about 1.x-1.x and 1.x-2.x updates as they are going to be deprecated soon, and the default for 2.x-2.x updates is that it should be allowed. 

Thus:

* device type needs not be checked
* updates for 2.x devices should be allowed in general
* the dev, downgrade, same version checks are good, and can stay
* The min target version for 2.x can be just 2.2.0+rev1 in general.

These checks are in the node SDK's modules, I understand but they are getting cleaned from there. The proxy does its checks as well, no need to repeat everything like this, IMHO we can let the tests be in the proxy alone (where it will be generally more permissive, as to no need to add new device types explicitly to add support for them in HUP, that's work in progress)

This part above probably mainly concerns `get_hup_action_type` further below in the module, I guess, but some parts this whole file.

Let me know if anything's strange about this mentioned above!

balena/settings.py
```diff
@@ -31,14 +31,19 @@ class Settings(object):
...
+                                'image_maker_endpoint', 'image_maker_endpoint_version',
+                                'device_actions_endpoint_version'])
 
     _setting = {
         # These are default config values to write default config file.
         # All values here must be in string format otherwise there will be error when write config file.
         'pine_endpoint': 'https://api.balena-cloud.com/v5/',
         'api_endpoint': 'https://api.balena-cloud.com/',
         'api_version': 'v5',
+        'image_maker_endpoint': 'https://img.balena-cloud.com/api/v1/',
```
imrehg, line 43: if all these are `.balena-cloud.com` for the endpoints, would it make sense to break it out to a single variable (the domain name), and use that for the endpoints with just the subdomains set there?

balena/models/hup.py
nghiant2710: Got it, I don't know much about how the hup proxy works so just tried to include every single check we have in the JS SDK. Updated!

balena/settings.py
nghiant2710: @imrehg I think these endpoints are essentially separate (for old SDK versions, they don't have same domain or in case of Open Balena, they can be anything) so I don't want to add unnecessary complexity here :)

balena/settings.py
```diff
@@ -31,14 +31,19 @@ class Settings(object):
...
+        'image_maker_endpoint_version': 'v1',
```
thgreasi: This seems not being used.
Should we remove the `v1` from the `image_maker_endpoint` and change the implementation or just drop this option?

balena/models/device_os.py
```diff
@@ -134,3 +139,95 @@ def parse_params(self, **params):
...
+            dict: the versions information, of the following structure:
+                * versions - an array of strings, containing exact version numbers supported by the current environment.
+                * recommended - the recommended version, i.e. the most recent version that is _not_ pre-release, can be `None`.
+                * latest - the most recent version, including pre-releases.
+                * default - recommended (if available) or latest otherwise.
+
+        """
+
+        response = self.base_request.request(
+            'image/{device_type}/versions'.format(device_type=device_type), 'GET',
```
thgreasi, line 220: Keep in mind that we will soon have to stop using this endpoint and use https://api.balena-cloud.com/device-types/v1/raspberrypi3/images/ instead.
HQ: https://github.com/balena-io/balena/issues/1744

Can you open an issue for this after this gets merged, so that we don't forget it & a reference appears in the HQ issue?

balena/models/device.py
```diff
@@ -1197,3 +1202,95 @@ def get_supervisor_state(self, uuid):
             '/supervisor/v1/device', 'POST', data=data,
             endpoint=self.settings.get('api_endpoint')
         )
+
+    def __check_os_update_target(self, device_info, target_os_version):
+        """
+        """
+
+        if 'uuid' not in device_info and not device_info['uuid']:
```
thgreasi: I think that this needs to be an `or`
```suggestion
        if 'uuid' not in device_info or not device_info['uuid']:
```

balena/models/device.py
```diff
@@ -1197,3 +1202,95 @@ def get_supervisor_state(self, uuid):
...
+            raise exceptions.OsUpdateError('The uuid of the device is not available')
+
+        if 'is_online' not in device_info and not device_info['is_online']:
```
thgreasi: If the above is true
```suggestion
        if 'is_online' not in device_info or not device_info['is_online']:
```

balena/models/device.py
```diff
@@ -1197,3 +1202,95 @@ def get_supervisor_state(self, uuid):
...
+            raise exceptions.OsUpdateError('The device is offline: {uuid}'.format(uuid=uuid))
+
+        if 'os_version' not in device_info and not device_info['os_version']:
```
thgreasi: If the above is true
```suggestion
        if 'os_version' not in device_info or not device_info['os_version']:
```

balena/models/device.py
```diff
@@ -1197,3 +1202,95 @@ def get_supervisor_state(self, uuid):
...
+            raise exceptions.OsUpdateError('The current os version of the device is not available: {uuid}'.format(uuid=uuid))
+
+        if 'device_type' not in device_info and not device_info['device_type']:
```
thgreasi: If the above is true
```suggestion
        if 'device_type' not in device_info or not device_info['device_type']:
```

balena/models/device.py
```diff
@@ -1197,3 +1202,95 @@ def get_supervisor_state(self, uuid):
...
+            >>> balena.models.device.start_os_update('b6070f4fea5edf808b576123157fe5ec', '2.29.2+rev1.prod')
+            {u'status': u'in_progress', u'action': u'resinhup', u'parameters': {u'target_version': u'2.29.2+rev1.prod'}, u'last_run': 1554490809219L}
+        """
+
+        device = self.get(uuid)
+        # this will throw an error if the action is not available
+        self.__check_os_update_target(device, target_os_version)
+
+        all_versions = self.device_os.get_supported_versions(device['device_type'])['versions']
+        if not [i for i in all_versions if semver.compare(target_os_version, i) == 0]:
```
thgreasi: NIT: How about `v` instead of `i` ?  :)
```suggestion
        if not [v for v in all_versions if semver.compare(target_os_version, v) == 0]:
```

thgreasi: Just as a heads up, we did have some releases that were not semver valid (eg: `2.0.0.rev1 (prod)`), and resin-semver is used to normalize all those weird cases. I'm not sure if we offer them any more (@imrehg can confirm probably), but maybe we could add a validity check as well in here, if the semver lib offers one:
```suggestion
        if not [i for i in all_versions if semver.valid(i) and semver.compare(target_os_version, i) == 0]:
```
or use the `__parse_balena_semver` which I just saw...

balena/models/device_os.py
```diff
@@ -134,3 +139,95 @@ def parse_params(self, **params):
             # if 'wifiKey' not in params:
             #    raise exceptions.MissingOption('wifiKey')
         return params
+
+    def __parse_balena_semver(self, os_version):
```
thgreasi: NIT: This sounds more like
```suggestion
    def __normalize_balena_semver(self, os_version):
```
(no strong opinion on this)

balena/models/device_os.py
```diff
@@ -134,3 +139,95 @@ def parse_params(self, **params):
...
+        """
+        safeSemver and trimOsText from resin-semver in Python.
+        ref: https://github.com/balena-io-modules/resin-semver/blob/master/src/index.js#L5-L24
+
+        """
+
+        # fix major.minor.patch.rev to use rev as build metadata
+        version = re.sub(r'\.rev', '+rev', os_version)
```
thgreasi: I think that this should be `/(\.[0-9]+)\.rev/` as in resin-sember, since otherwise `Resin OS 2.0.1+prod.rev1` would end up with two `+`, which would be invalid.
PS: Except if we handle that elsewhere.

balena/models/device_os.py
```diff
@@ -134,3 +139,95 @@ def parse_params(self, **params):
...
+        # fix major.minor.patch.prod to be treat .dev & .prod as build metadata
+        version = re.sub(r'\.(dev|prod)', r'+\1', version)
```
thgreasi: Again we might end up with more than one `+`. Eg: `Resin OS 2.0.1+rev1.dev`

balena/models/device_os.py
```diff
@@ -134,3 +139,95 @@ def parse_params(self, **params):
...
+            'image/{device_type}/versions'.format(device_type=device_type), 'GET',
+            endpoint=self.settings.get('image_maker_endpoint'), auth=False
+        )
+
+        potential_recommended_versions = [i for i in response['versions'] if not re.search(r'(\.|\+|-)dev', i)]
+        potential_recommended_versions = [i for i in potential_recommended_versions if not semver.parse(i)['prerelease']]
+        recommended = potential_recommended_versions[0] if potential_recommended_versions else None
+
+        return {
+            'versions': response['versions'],
```
thgreasi, line 229: Just a reminder that the js SDK also does a sorting so that he returned versions are returned in a preference order, where it sorts by version, then by rev then by dev/prod status. I'm not sure whether the python sdk should also do that, and the js sdk is doing it so that the dashboard & cli are presenting the versions in the "preferred - stable" order.

balena/models/hup.py
```diff
@@ -0,0 +1,42 @@
...
+
+        xstr = lambda s: '' if s is None else str(s)
+        if 'dev' in xstr(parsed_current_ver['prerelease']) + xstr(parsed_target_ver['prerelease']) + xstr(parsed_target_ver['build']) + xstr(parsed_current_ver['build']):
+            raise exceptions.OsUpdateError('Updates cannot be performed on development balenaOS variants')
+
+        if semver.compare(target_version, current_version) < 0:
+            raise exceptions.OsUpdateError('OS downgrades are not allowed')
+
+        # For 1.x -> 2.x or 2.x to 2.x only
+        if parsed_target_ver['major'] > 1 and semver.compare(target_version, self.MIN_TARGET_VERSION) < 0:
```
thgreasi, line 39: @imrehg Would it make sense to check that `parsed_target_ver['major'] == 2` so that we don't allow v2 -> v3 upgrades right away? In other words, I would prefer that the target version satisfies `^2.2.0+rev1`, which isn't exactly what `semver.compare` does.

thgreasi, line 39: Since we don't include any of the v1 checks, would it make sense to throw an error when the `current_version` or the `target_version` are v1?

thgreasi (commented):
Great work @nghiant2710 :+1:

balena/settings.py
nghiant2710: Dropped this option.

balena/models/device_os.py
nghiant2710: @thgreasi thanks for the heads up, created an issue!

balena/models/device.py
nghiant2710: updated

balena/models/device.py
nghiant2710: @thgreasi indeed, __normalize_balena_semver (__parse_balena_semver renamed) handles this.

balena/models/device_os.py
nghiant2710: renamed

balena/models/device_os.py
nghiant2710: ah right, updated

balena/models/device_os.py
nghiant2710: updated

balena/models/hup.py
nghiant2710: @thgreasi, as @imrehg mentioned in this thread https://www.flowdock.com/app/rulemotion/resin-frontend/threads/BtMD-GY0G565-Rlyirp6vbzmcEk, I think we should leave this case for the proxy to handle.

nghiant2710: same as above, I think we should leave this for the proxy to handle

nghiant2710:
@thgreasi, thanks for the comments, I've fixed them all, can you please check again?

balena/settings.py
```diff
@@ -31,14 +31,18 @@ class Settings(object):
     CONFIG_FILENAME = 'balena.cfg'
     DEFAULT_SETTING_KEYS = set(['pine_endpoint', 'api_endpoint', 'api_version',
                                 'data_directory', 'image_cache_time',
-                                'token_refresh_interval', 'cache_directory', 'timeout'])
+                                'token_refresh_interval', 'cache_directory', 'timeout',
+                                'image_maker_endpoint', 'image_maker_endpoint_version',
```
thgreasi: Forgotten `image_maker_endpoint_version` occurrence
```suggestion
                                'image_maker_endpoint',
```

balena/models/hup.py
thgreasi: Cool, let's see what @imrehg thinks on this :+1:

thgreasi (approved):
LGTM :+1: 
Just remove the extra occurrence of `image_maker_endpoint_version` & let's see what @imrehg thinks.

tests/pep8-git-pr-checker
```diff
@@ -7,7 +7,7 @@ import sys
 import tempfile
 
 # 501: line too long
```
xginn8, line 9: is it worth adding `E731` here for accessibility?

balena/models/hup.py
imrehg: @thgreasi yeah, have the proxy to handle, and going forward it should be all supported.

imrehg: Yeah, same as above.

imrehg (approved):
LGTM, thanks a bunch @nghiant2710 !

PR #134, opened 2019-04-09
device: Add support for get_supervisor_target_state device method

xginn8:
Connects-to: https://github.com/balena-io/balena-sdk/pull/653
Change-type: minor
Signed-off-by: Matthew McGinn <mamcgi@gmail.com>


tests/functional/models/test_device.py
```diff
@@ -374,6 +374,17 @@ def test_get_supervisor_state(self):
             self.balena.models.device.get_supervisor_state(device['uuid'])
         self.assertIn('No online device(s) found', cm.exception.message)
 
+    def test_get_supervisor_target_state(self):
+        device = self.helper.create_device()[1]
+
+        # should be rejected if the device doesn't exist.
```
nghiant2710: @xginn8, this test case isn't correct, I've tried and the API will return a 304 with no message-body when try to GET the state of invalid uuid or uuid you don't have access. I think  you should GET the device first in the code with `self.get(uuid)` which will raise `DeviceNotFound` if you don't have access to that uuid and update this case to catch that exception.

tests/functional/models/test_device.py
```diff
@@ -374,6 +374,17 @@ def test_get_supervisor_state(self):
...
+        with self.assertRaises(Exception) as cm:
+            self.balena.models.device.get_supervisor_target_state('9999999')
+        self.assertIn('No online device(s) found', cm.exception.message)
+        # should be rejected if the device doesn't exist.
```
nghiant2710: @xginn8, I think this case is about offline device, not the one doesn't exist. Also I'm able to GET the state of offline device so we shouldn't use `assertRaises` here

nghiant2710:
hey @xginn8, can you please check few comments here?

xginn8:
> hey @xginn8, can you please check few comments here?

sorry @nghiant2710 I missed your note somehow. I am fixing up the tests now and should be ready for review shortly!

xginn8:
@nghiant2710 updated, PTAL!

balena/models/device.py
```diff
@@ -1197,3 +1197,24 @@ def get_supervisor_state(self, uuid):
...
+            uuid (str): device uuid.
+
+        Returns:
+            dict: supervisor target state.
+
+        Examples:
+            >>> balena.models.device.get_supervisor_target_state('b6070f4fea5edf808b576123157fe5ec')
+            {u'local': {u'name': u'holy-darkness', u'config': {u'RESIN_SUPERVISOR_NATIVE_LOGGER': u'true', u'RESIN_SUPERVISOR_POLL_INTERVAL': u'900000'}, u'apps': {u'1398898': {u'name': u'test-nuc', u'commit': u'f9d139b80a7df94f90d7b9098b1353b14ca31b85', 6 u'releaseId': 850293, u'services': {u'229592': {u'imageId': 1016025, u'serviceName': u'main', u'image': u'registry2.balena-cloud.com/v2/27aa30131b770a4f993da9a54eca6ed8@sha256:f489c30335a0036ecf1606df3150907b32ea39d73ec6de825a549385022e3e22', u'running': True, u'environment': {}, u'labels': {u'io.resin.features.dbus': u'1', u'io.resin.features.firmware': u'1', u'io.resin.features.kernel-modules': u'1', u'io.resin.features.resin-api': u'1', u'io.resin.features.supervisor-api': u'1'}, u'privileged': True, u'tty': True, u'restart': u'always', u'network_mode': u'host', u'volumes': ['resin-data:/data']}}, 6 u'volumes': {u'resin-data': {}}, u'networks': {}}}}, u'dependent': {u'apps': {}, u'devices': {}}}
```
thgreasi, line 1213: I'm not sure what `6 u'releaseId'` & `6 u'volumes'` means :thinking:

thgreasi (approved):
LGTM

PR #135, opened 2019-04-12
added function to set device to specific build_id

amirfuhrmann:


nghiant2710:
@amirfuhrmann, thank you very much for your contribution. I've fixed the `device.set_to_release()` method to accept only `success` release so you won't ever have the reported issue again. However, I will also add your method to the SDK  to provide a fast way for experienced user like you to do this task.

I will merge your PR into my working branch, do some cleanup and add you as a reviewer.
Thank you!

PR #137, opened 2019-04-12
device: replace 'last_seen_time' with 'last_vpn_event', add unit test

xginn8:
Change-type: patch
Signed-off-by: Matthew McGinn <matthew@balena.io>


nghiant2710 (approved):

nghiant2710:
@xginn8, thanks for the PR 👍 I don't have my laptop with me so can't check few issues raised from support but will do as soon as I'm back at my desk

PR #139, opened 2019-04-13
serviceEnvVar: Fix bug in service_env_var.create()

### LINKED ISSUE #131: RequestError in balena.models.environment_variables.service_environment_variable.create
I am getting a RequestError exception when using RequestError when using balena.models.environment_variables.service_environment_variable.create 

I believe it is due to the list comprehension in 

https://github.com/balena-io/balena-sdk-python/blob/0a28e3f349572664e871dfb40aef8ee2e543bf50/balena/models/environment_variables.py#L543

and the list is passed to the data variable

```
        data = {
            'name': env_var_name,
            'value': value,
            'service': service_id
        }
```

https://github.com/balena-io/balena-sdk-python/blob/0a28e3f349572664e871dfb40aef8ee2e543bf50/balena/models/environment_variables.py#L548

I have tried 
```
        data = {
            'name': env_var_name,
            'value': value,
            'service': service_id[0]
        }
```
and it works. 

I suggest to also add a check to the list as in 

https://github.com/balena-io/balena-sdk-python/blob/0a28e3f349572664e871dfb40aef8ee2e543bf50/balena/models/environment_variables.py#L247

nghiant2710:
Fixes #131 

Change-type: patch


nghiant2710:
ping @thgreasi, @shaunmulligan, @xginn8

thgreasi (approved):
I would prefer an early return pattern, but LGTM :+1: 
```
if not service_id:
    raise exceptions.ServiceNotFound(service_name)
# no else & no extra indentation
data = {...
```

PR #146, opened 2019-05-03
Fix POST parameter name

eugenefab:
In the `create` method the POST parameter for the environment variable name is currently incorrect. It's changed to the value described in the API documentation: `name`.

### DIFF
--- a/balena/models/environment_variables.py
+++ b/balena/models/environment_variables.py
@@ -97,7 +97,7 @@ class DeviceEnvVariable(object):
         device = self.device.get(uuid)
         data = {
             'device': device['id'],
-            'env_var_name': env_var_name,
+            'name': env_var_name,
             'value': value
         }
         new_env_var = json.loads(self.base_request.request(


xginn8 (approved):
Thanks for contributing @eugenefab! This change lgtm. It would be good to validate these end-to-end by perhaps spinning up the API from `open-balena` or something, but that's outside the scope of this change. I want to wait for @nghiant2710's feedback before merging since the tests are broken and I am not sure how external contributions will fit into our versioning system.

eugenefab:
> Thanks for contributing @eugenefab! This change lgtm. It would be good to validate these end-to-end by perhaps spinning up the API from `open-balena` or something, but that's outside the scope of this change. I want to wait for @nghiant2710's feedback before merging since the tests are broken and I am not sure how external contributions will fit into our versioning system.

@xginn8 Yea, I was trying to see why the tests were failing and how to fix them, but I'm not seeing why this change would cause a failure. Seems almost like a test configuration issue. If it helps, I just deployed a fork of this SDK with this change in it, to our own integration environment and the API calls are succeeding now.

Thanks for looking at this so quickly!

nghiant2710:
@eugenefab thank you for the PR, the tests failed since our CI can't pull that branch from your fork. LGTM but I will need to do a review if there are other API changes that break the SDK so I will merge this to my working branch and will release really soon.

PR #147, opened 2019-05-05
env_var: Fix POST parameter name

### LINKED ISSUE #145: Creating environment variables fails with 'Database Error'
In the past week, attempts to call DeviceEnvVariable.create() have failed with `Database Error` which appears to come back as a 404 from your API. I've modified the source here:
https://github.com/balena-io/balena-sdk-python/blob/master/balena/models/environment_variables.py#L100
by changing the parameter name to just 'name' from 'env_var_name' and the call started working correctly again. Since the above code has been the same for many versions of this SDK, I suspect something is going on the API server side of things?

This is a blocker for our code base. Thoughts on how to proceed would be appreciated!

nghiant2710:
Connects to https://github.com/balena-io/balena-sdk-python/pull/146, Fixes #145 

In the `create` method the POST parameter for the environment variable name is currently incorrect. It's changed to the value described in the API documentation: `name`.

Change-type: patch

### DIFF
--- a/balena/models/environment_variables.py
+++ b/balena/models/environment_variables.py
@@ -97,7 +97,7 @@ class DeviceEnvVariable(object):
         device = self.device.get(uuid)
         data = {
             'device': device['id'],
-            'env_var_name': env_var_name,
+            'name': env_var_name,
             'value': value
         }
         new_env_var = json.loads(self.base_request.request(


nghiant2710:
cc @eugenefab, I updated the commit message a bit so it can pass our checks. Thank you again for the PR 👍

xginn8 (approved):
lgtm!

PR #151, opened 2019-06-10
device: Update get_status method based on latest JS SDK changes

nghiant2710:
Connects-to: https://github.com/balena-io/balena-sdk/pull/641
Change-type: minor

### DIFF
--- a/balena/models/device.py
+++ b/balena/models/device.py
@@ -33,6 +33,7 @@ class DeviceStatus(object):
     UPDATING = "Updating"
     OFFLINE = "Offline"
     POST_PROVISIONING = "Post Provisioning"
+    INACTIVE = "Inactive"
 
 
 class Device(object):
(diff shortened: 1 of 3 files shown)


balena/models/device.py
```diff
@@ -835,11 +836,14 @@ def get_status(self, uuid):
 
         """
 
-        device = self.get(uuid)
+        device = self.get_with_service_details(uuid)
+        if not device['is_active']:
+            return DeviceStatus.INACTIVE
```
xginn8, line 841: I think this new status should be added to the docstring as well (line 831)

balena/models/device.py
```diff
@@ -852,6 +856,13 @@ def get_status(self, uuid):
         if device['download_progress'] is not None:
             return DeviceStatus.CONFIGURING
 
+        if device['current_services']:
+            for service_name in device['current_services']:
```
xginn8, line 860: if you loop through `for service_name, installs in device['current_services'].items()` here, you can drop the separate assignment in line 861.

xginn8 (approved):
some small nits, otherwise lgtm!

balena/models/device.py
nghiant2710: @xginn8, updated. Thanks!

balena/models/device.py
nghiant2710: @xginn8, `dict.items()` is a bit different between Python 2.x and 3.x. Since our code should be compatible with both so I think it's better to implement it like that.

xginn8: @nghiant2710 fwiw I had tested this behavior in both py2/3 before recommending the change (it is compatible), but if you prefer this implementation that's fine with me.

PR #154, opened 2019-08-12
release: Support retrieving releases & managing release tags by commit

nghiant2710:
Connects-to: https://github.com/balena-io/balena-sdk/pull/725
Change-type: major

### DIFF
--- a/balena/models/release.py
+++ b/balena/models/release.py
@@ -5,6 +5,7 @@ from ..settings import Settings
 from .. import exceptions
 from .image import Image
 from .service import Service
+from ..utils import is_id
 
 
 class Release(object):
@@ -73,12 +74,12 @@ class Release(object):
         else:
             raise exceptions.ReleaseNotFound(raw_query)
 
-    def get(self, id):
+    def get(self, commit_or_id):
         """
         Get a specific release.
 
         Args:
-            id (str): release id.
+            commit_or_id: release commit (str) or id (int).
 
         Returns:
             dict: release info.
@@ -88,7 +89,20 @@ class Release(object):
 
         """
 
-        return self.__get_by_option(id=id)[0]
+        if is_id(commit_or_id):
+            return self.__get_by_option(id=commit_or_id)[0]
+        else:
+            raw_query = "$filter=startswith(commit, '{}')".format(commit_or_id)
+
+            try:
+                rt = self.__get_by_raw_query(raw_query)
+
+                if len(rt) > 1:
+                    raise exceptions.AmbiguousRelease(commit_or_id)
+
+                return rt[0]
+            except exceptions.ReleaseNotFound:
+                raise exceptions.ReleaseNotFound(commit_or_id)
 
     def get_all_by_application(self, app_id):
         """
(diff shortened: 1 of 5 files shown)


balena/utils.py
```diff
@@ -0,0 +1,15 @@
+import numbers
+
+
+def is_id(value):
+    """
+    Return True, if the input value is a valid ID. False otherwise.
```
xginn8, line 6: this function just checks if the input could be an integer, not necessarily that it's a valid id, right?

xginn8 (approved):

balena/models/release.py
```diff
@@ -137,6 +151,8 @@ def get_with_image_details(self, id):
 
         """
 
+        id = self.get(commit_or_id)['id']
```
thgreasi, line 154: There is a `raise exceptions.ReleaseNotFound(id)` a bit lower, which while it's dead code atm (can't be reached b/c `self.get` has already checked that the release exists) could be changed to `raise exceptions.ReleaseNotFound(commit_or_id)`.

balena/models/tag.py
```diff
@@ -339,23 +341,25 @@ def get_all_by_application(self, app_id):
 
         return super(ReleaseTag, self).get_all(raw_query=query)
 
-    def get_all_by_release(self, release_id):
+    def get_all_by_release(self, commit_or_id):
```
thgreasi, line 344: Should we now mention that this also throws an AmbiguousReleaseError?

balena/models/tag.py
```diff
@@ -378,12 +382,12 @@ def get_all(self):
 
         return super(ReleaseTag, self).get_all()
 
-    def set(self, release_id, tag_key, value):
+    def set(self, commit_or_id, tag_key, value):
```
thgreasi, line 385: > Should we now mention that this also throws an AmbiguousReleaseError?

balena/models/tag.py
```diff
@@ -392,27 +396,30 @@ def set(self, release_id, tag_key, value):
...
             tag_key (str): tag key.
 
         Examples:
-            >>> balena.models.tag.release.remove('135', 'releaseTag1')
+            >>> balena.models.tag.release.remove(135, 'releaseTag1')
             OK
 
         """
 
+        release_id = self.release.get(commit_or_id)['id']
```
thgreasi, line 424: > Should we now mention that this also throws an AmbiguousReleaseError?

Also, did this throw 404's before this change when called w/ a random id?
PS: It might even make sense to use `is_id` to avoid the extra request, but whether we can use it or not depends on the above question.

balena/models/release.py
nghiant2710: updated!

balena/models/tag.py
nghiant2710: @thgreasi, `release.get` does use `is_id` to check the parameter value first and throw AmbiguousRelease and ReleaseNotFound exception so we don't need to repeat that in the above tag methods

thgreasi: Oh ok, didn't know that it's inferred automatically :+1:

PR #161, opened 2019-10-14
device: Add local mode related methods

nghiant2710:
See: https://github.com/balena-io/balena-sdk/pull/733

Change-type: minor

### DIFF
--- a/balena/models/device.py
+++ b/balena/models/device.py
@@ -17,9 +17,15 @@ from .device_os import DeviceOs
 from ..settings import Settings
 from ..auth import Auth
 from .. import exceptions
+from ..resources import Message
 from .application import Application
 from .release import Release
 from .hup import Hup
+
+
+LOCAL_MODE_MIN_OS_VERSION = '2.0.0'
+LOCAL_MODE_MIN_SUPERVISOR_VERSION = '4.0.0'
+LOCAL_MODE_ENV_VAR = 'RESIN_SUPERVISOR_LOCAL_MODE'
 
 
 # TODO: support both device uuid and device id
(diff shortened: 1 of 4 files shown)


balena/models/device.py
```diff
@@ -1361,3 +1367,158 @@ def get_os_update_status(self, uuid):
...
+            raise exceptions.LocalModeError(Message.DEVICE_SUPERVISOR_NOT_SUPPORT_LOCAL_MODE)
+
+        if device['os_variant'] != 'dev':
+            raise exceptions.LocalModeError(Message.DEVICE_OS_TYPE_NOT_SUPPORT_LOCAL_MODE)
+
+    def __upsert_device_local_mode(self, device, value):
+        try:
+            data = {
+                'device': device['id'],
+                'application': device['belongs_to__application']['__id'],
```
thgreasi: I think that the `application` id isn't necessary and the `device` id, along with the `name` & `value` is all that's needed.
Even if it works today, in the future it could throw a BadRequest b/c of the extra unknown properties.

balena/models/device.py
```diff
@@ -1361,3 +1367,158 @@ def get_os_update_status(self, uuid):
...
+        }
+
+        response = self.base_request.request(
+            'device_config_variable', 'GET', params=params,
+            endpoint=self.settings.get('pine_endpoint')
+        )['d']
+
+        if len(response) == 0:
+            return False
+        if response[0]['value'] != '1':
```
thgreasi: Optional-NIT: Imo a `== 1` `return True` check is more verbose and follows more the 
> local mode is enabled iff the value equals `1`

PS: I'm not saying it's wrong and I'm fine with this as is.

thgreasi (commented):
Other than the redundant (afaik) `application` id passed in the `device_config_variable`POST, LGTM :+1:

balena/models/device.py
nghiant2710: Updated!

balena/models/device.py
nghiant2710: Updated!

nghiant2710:
@thgreasi I've fixed all your comments, can you please have a look?

thgreasi (approved):

Issue #163, opened 2019-11-06
Bug in the `application.get_all()` & `device.get_all()` methods


richbayliss:
User reports that the `application.get_all()` method returns an empty array for them. @thgreasi  was able to replicate the issue by creating a simple balena app with the python sdk & authenticated using both the injected `BALENA_API_KEY` as well as a new user API key.

In both cases `application.get_all()` returned an empty array, `application.get(<app_name>)` was only able to retrieve the current app & `device.get_all()` was only returning the current device.

xginn8:
@thgreasi using a device api key in the node.js SDK exhibits the same behavior (access to only the device who's key it is, and no application access [empty array]). is it possible that this behavior is actually correct?

thgreasi:
@xginn8 the results are correct given the device api key, the fact that the user can't set a different api key to be used by the sdk when the `BALENA_API_KEY` or `RESIN_API_KEY` env var exists is the actual UX issue.
I would expect that calling `login_with_token()` should allow me to specify in a different api key that I want the SDK to be using from that point and after. That's what the user (& myself) was trying to do.

nghiant2710:
Fixed in https://github.com/balena-io/balena-sdk-python/pull/169

PR #165, opened 2019-11-07
auth: allow logins if BALENA/RESIN_API_KEY is already defined

xginn8:
If BALENA/RESIN_API_KEY is defined (potentially in the environment),
subsequent login attempts are unsuccessful since we always default to
that key.

Connects-to: #163
Change-type: minor
Signed-off-by: Matthew McGinn <matthew@balena.io>


nghiant2710:
@xginn8, if the API_KEY is found then the SDK will always use it instead of auth token for authentication https://github.com/balena-io/balena-sdk-python/blob/5872c66075d54afd0a30a163cf2f492135e8b40f/balena/base_request.py#L140-L141 so I'm not sure this PR will fix the issue.

Actually, we haven't changed this behavior for years so users just hit this bug recently then something changed in the API or at least with the permissions of API_KEY since it no longer has access to expected resources. cc @thgreasi

thgreasi:
This sounds a bit strange @nghiant2710 :thinking: 
In other words, when the SDK runs on an environment that has the BALENA_API_KEY variable set, any calls that the user does on `auth.login()` and `auth.login_with_token()` have no effect.
Imo it makes more sense to
* check whether the setting have an api key and use that
* otherwise check whether an env var with an api key is preset

xginn8:
@nghiant2710 by setting `login=True`, we are able to override those headers: https://github.com/balena-io/balena-sdk-python/blob/5872c66075d54afd0a30a163cf2f492135e8b40f/balena/base_request.py#L154-L157

You can confirm this patch locally using the following example:

```
❯ cat test.py 
#!/usr/bin/env python3

import os
from balena import Balena

balena = Balena()
balena.settings.set('token', '')
balena.settings.set('cache_directory', '')
balena.settings.set('data_directory', '/tmp')

balena.auth.login_with_token(os.environ['TOKEN'])
print(balena.auth.get_email())
```

```
~/balena/balena-sdk-python 163-login-with-token
❯ BALENA_API_KEY=1 TOKEN=$(cat ~/dls/Api\ key\ baletheus\ 2019-05-09\ 09_24_17) python ./test.py
mamcgi@protonmail.com
```

```
~/balena/balena-sdk-python master
❯ BALENA_API_KEY=1 TOKEN=$(cat ~/dls/Api\ key\ baletheus\ 2019-05-09\ 09_24_17) python ./test.py
  File "./test.py", line 12, in <module>
    print(balena.auth.get_email())
  File "/home/matthew/balena/balena-sdk-python/balena/auth.py", line 237, in get_email
    return self.__get_property('email')
  File "/home/matthew/balena/balena-sdk-python/balena/auth.py", line 56, in __get_property
    if element in self.__get_user_data():
  File "/home/matthew/balena/balena-sdk-python/balena/auth.py", line 35, in __get_user_data
  File "/home/matthew/balena/balena-sdk-python/balena/base_request.py", line 191, in request
    raise exceptions.RequestError(response._content)
balena.exceptions.RequestError: b'Unauthorized'
```

nghiant2710:
@xginn8, it works because you passed `login=True`, others will still fail because the SDK will use the API_KEY instead of auth token if API_KEY is found when making the request. So I don't think we should fix this by adding that option to all other methods.

It's better to fix as @thgreasi mentioned above, we should always prioritize the auth token and only looks for the API_KEY when the auth token is not set. I would be happy to update the PR if you still have other things on your plate.

PR #168, opened 2019-11-13
auth: Prioritize auth token over API_KEY

### LINKED ISSUE #163: Bug in the `application.get_all()` & `device.get_all()` methods

nghiant2710:
Connects-to: https://github.com/balena-io/balena-sdk-python/pull/165
Fixes #163 

If auth token found, the SDK will always use it even if BALENA_API_KEY/RESIN_API_KEY is set

Change-type: minor

### DIFF
--- a/balena/base_request.py
+++ b/balena/base_request.py
@@ -156,18 +156,26 @@ class BaseRequest(object):
             api_key = None
             auth = True
 
-        if auth and not api_key:
-            if not self.settings.has(TOKEN_KEY):
+        if auth:
+            if self.settings.has(TOKEN_KEY):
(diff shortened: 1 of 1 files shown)


thgreasi (approved):
LGTM :+1: 
I guess that `settings.has(TOKEN_KEY)` does check that there is indeed a value there, and returns `false` if there is the key but there is no value in the store.

PR #169, opened 2019-11-19
Release balena-sdk for Python v9.0.0

nghiant2710:
All separate commits of this PR were individually approved and this is just a meta-PR for releasing v12 so that the CHANGELOG properly communicates the changes.
A last check is welcome of course.

### DIFF
--- a/balena/models/tag.py
+++ b/balena/models/tag.py
@@ -2,6 +2,7 @@ import json
 
 from ..base_request import BaseRequest
 from .device import Device
+from .release import Release
 from ..settings import Settings
 from .. import exceptions
 
@@ -317,6 +318,7 @@ class ReleaseTag(BaseTag):
 
     def __init__(self):
         super(ReleaseTag, self).__init__('release')
+        self.release = Release()
 
     def get_all_by_application(self, app_id):
         """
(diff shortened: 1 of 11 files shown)


balena/models/tag.py
```diff
@@ -339,23 +341,25 @@ def get_all_by_application(self, app_id):
...
         Args:
-            release_id (str): release id.
+            commit_or_id: release commit (str) or id (int).
 
         Returns:
             list: list contains release tags.
 
         Examples:
-            >>> balena.models.tag.release.get_all_by_release('135')
+            >>> balena.models.tag.release.get_all_by_release(135)
```
thgreasi, line 355: Based on the example result, this should be
```suggestion
            >>> balena.models.tag.release.get_all_by_release(465307)
```
Could we also have an example using the commit hash as an argument?

balena/models/tag.py
```diff
@@ -392,27 +396,30 @@ def set(self, release_id, tag_key, value):
...
         Remove a release tag.
 
         Args:
-            release_id (str): release id.
+            commit_or_id: release commit (str) or id (int).
             tag_key (str): tag key.
 
         Examples:
-            >>> balena.models.tag.release.remove('135', 'releaseTag1')
+            >>> balena.models.tag.release.remove(135, 'releaseTag1')
```
thgreasi, line 419: NIT: Let's change this to `465307` so that it's consistent with the above examples.

thgreasi (approved):
Just three NITs on the docs, but other than that LGTM :+1:

PR #170, opened 2019-11-20
auth: Stop sending api key in query string

nghiant2710:
The SDK should never send the api key as query string and only via http headers

Connects-to: https://github.com/balena-io/balena-sdk/pull/780

Change-type: major

### DIFF
--- a/balena/base_request.py
+++ b/balena/base_request.py
@@ -37,9 +37,11 @@ class BaseRequest(object):
     def __set_content_type(self, headers, ctype):
         headers.update({'Content-Type': ctype})
 
-    def __set_authorization(self, headers):
-        headers.update(
-            {'Authorization': 'Bearer {:s}'.format(self.settings.get(TOKEN_KEY))})
+    def __set_authorization(self, headers, key=None):
+        if not key:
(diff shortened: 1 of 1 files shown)


afitzek (approved):
lgtm

balena/base_request.py
```diff
@@ -137,11 +137,11 @@ def __request(self, url, method, params, endpoint, headers=None,
...
         if auth:
-            if not api_key:
-                self.__set_authorization(headers)
-            params = self._format_params(params, api_key=api_key, raw_query=raw_query)
+            # api_key will be None if Auth token is found so we can just pass api_key here.
+            self.__set_authorization(headers, api_key)
+            params = self._format_params(params, raw_query=raw_query)
         else:
-            params = self._format_params(params, api_key=None, raw_query=raw_query)
+            params = self._format_params(params, raw_query=raw_query)
```
thgreasi: Let's move this line outside of the `if-else` and drop the `else`.
```suggestion
            params = self._format_params(params, raw_query=raw_query)
```

thgreasi (approved):

balena/base_request.py
nghiant2710: Updated!

thgreasi:
Already approved :+1:

PR #174, opened 2020-01-07
Fix/Add OS dowload methods

### LINKED ISSUE #173: Unable to download BalenaOS Images using PythonSDK
Using the Balena Python SDK (version 9.0.0) to download balenaOS images and im getting this error,

  File "test_download.py", line 6, in <module>
    r = balena.models.device_os.download(**data)
  File "/home/eis/.local/lib/python2.7/site-packages/balena/models/device_os.py", line 102, in download
    endpoint=self.settings.get('api_endpoint'), stream=True, login=True
  File "/home/eis/.local/lib/python2.7/site-packages/balena/base_request.py", line 182, in request
    stream=stream, auth=auth, api_key=api_key, raw_query=raw_query)
  File "/home/eis/.local/lib/python2.7/site-packages/balena/base_request.py", line 145, in __request
    return request_method(url, headers=headers, data=data, stream=stream)
  File "/home/eis/.local/lib/python2.7/site-packages/balena/base_request.py", line 54, in __post
    url, data=json.dumps(self.util.decode_utf8(data)), headers=headers, stream=stream, timeout=self.timeout)
  File "/home/eis/.local/lib/python2.7/site-packages/requests/api.py", line 116, in post
    return request('post', url, data=data, json=json, **kwargs)
  File "/home/eis/.local/lib/python2.7/site-packages/requests/api.py", line 60, in request
    return session.request(method=method, url=url, **kwargs)
  File "/home/eis/.local/lib/python2.7/site-packages/requests/sessions.py", line 533, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/eis/.local/lib/python2.7/site-packages/requests/sessions.py", line 646, in send
    r = adapter.send(request, **kwargs)
  File "/home/eis/.local/lib/python2.7/site-packages/requests/adapters.py", line 529, in send
    raise ReadTimeout(e, request=request)
requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='api.balena-cloud.com', port=443): Read timed out. (read timeout=30.0)

python code:

```python
from balena import Balena                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
balena = Balena()                                                                                                                                                                                                                                                         
balena.auth.login_with_token('<access_token>')    
data = {'appId': '<app_id>', 'version': 'balenaOS 2.41.1+rev1', 'network': 'ethernet'}
r = balena.models.device_os.download(**data)
with open('balena.img', 'wb') as save_file:
    save_file.write(r.content)
```

nghiant2710:
- Fix device_os.download() method.
- Add device_os.download_unconfigured_image() method.

Fixes #173


balena/models/device_os.py
```diff
@@ -68,14 +68,21 @@ def get_config(self, app_id, options):
...
+    def download(self, raw=False, is_dev_image=False, **data):
         """
         Download an OS image. This function only works if you log in using credentials or Auth Token.
 
         Args:
             raw (bool): determining function return value.
+            is_dev_image (Optional[bool]): select development image instead of production image. defaults to False.
             **data: os parameters keyword arguments.
-                Details about os parameters can be found in parse_params function
+                version (str): the Balena OS version of the image. The SDK will try to parse version into semver-compatible version, unsupported (unpublished) version will result in rejection.
```
thgreasi: I think this should be `balenaOS`, as per our site
See: https://www.balena.io/os/

Also how about:
```suggestion
                version (str): the balenaOS version of the image. The SDK will try to parse the provided version as a semver-compatible version. Providing unavailable versions will result a rejection.
```

balena/models/device_os.py
```diff
@@ -68,14 +68,21 @@ def get_config(self, app_id, options):
             endpoint=self.settings.get('api_endpoint')
         )
 
-    def download(self, raw=None, **data):
+    def download(self, raw=False, is_dev_image=False, **data):
```
thgreasi: Doesn't that mean that if someone so far was `2.41.1+rev1.dev`, then they will now get a prod image? Maybe we can infer the dev variant with a `/\bdev\b/` regex.

balena/models/device_os.py
```diff
@@ -68,14 +68,21 @@ def get_config(self, app_id, options):
             endpoint=self.settings.get('api_endpoint')
         )
 
-    def download(self, raw=None, **data):
```
thgreasi, line 71: Was this working w/o arguments before (picking the latest as a default)?

balena/models/device_os.py
```diff
@@ -107,6 +119,49 @@ def download(self, raw=None, **data):
         else:
             return response
 
+    def download_unconfigured_image(self, device_type, version='latest', raw=False):
+        """
+        Download an unconfigured OS image. This function only works if you log in using credentials or Auth Token.
```
thgreasi: I just wanted to clarify that the API allows listing the versions & downloading unconfigured images w/o requiring authentication.
See: https://github.com/balena-io/balena-api/blob/8c3151df679357a5783b1142cdb0901f778ef8a4/src/routes/index.ts#L208-L209

So technically we could allow users do so w/o requiring authentication.

balena/models/device_os.py
```diff
@@ -107,6 +119,49 @@ def download(self, raw=None, **data):
...
+            >>> response['headers']
+            >>> response.headers
+            {'Access-Control-Allow-Headers': 'Content-Type, Authorization, Application-Record-Count, MaxDataServiceVersion, X-Requested-With, X-Balena-Client', 'content-disposition': 'attachment; filename="balena-cloud-raspberry-pi2-2.43.0+rev1-v10.2.2.img"', 'X-Content-Type-Options': 'nosniff', 'Access-Control-Max-Age': '86400', 'Transfer-Encoding': 'chunked', 'x-powered-by': 'Express', 'content-encoding': 'gzip', 'x-transfer-length': '134445269', 'last-modified': 'Mon, 23 Sep 2019 15:21:33 GMT', 'Connection': 'keep-alive', 'Access-Control-Allow-Credentials': 'true', 'Date': 'Tue, 07 Jan 2020 18:14:47 GMT', 'X-Frame-Options': 'DENY', 'Access-Control-Allow-Methods': 'GET, PUT, POST, PATCH, DELETE, OPTIONS, HEAD', 'Content-Type': 'application/octet-stream', 'Access-Control-Allow-Origin': '*'}
+
+        """
+
+        if version == 'latest':
+            version = self.get_supported_versions(device_type)['latest']
+        else:
+            version = self.get_device_os_semver_with_variant(version, 'prod')
```
thgreasi: I think that we also support downloading uncofigured dev images. As suggested above, how about inferring the `dev`/`prod` variant using a `/\bdev\b/` regex?

balena/models/device_os.py
nghiant2710: updated!

balena/models/device_os.py
nghiant2710: Nope, they will get a dev image, actually I didn't expect they will pass version name with `.prod` or `.dev` (since I don't see that from the version on dashboard) so I add `is_dev_image` so they can get dev image if needed but looks like we don't need it here as we assume user will pass os_variant in version.

Removed that option and I will leave version as it is. (The SDK already has the regex to handle dev variant here https://github.com/balena-io/balena-sdk-python/blob/master/balena/models/device_os.py#L155-L169)

balena/models/device_os.py
nghiant2710: version wasn't mandatory and it's parsed properly so this didn't work

balena/models/device_os.py
nghiant2710: Updated

balena/models/device_os.py
nghiant2710: Updated (The SDK already has the regex to handle dev variant here https://github.com/balena-io/balena-sdk-python/blob/master/balena/models/device_os.py#L155-L169)

balena/models/device_os.py
```diff
@@ -107,6 +116,49 @@ def download(self, raw=None, **data):
         else:
             return response
 
+    def download_unconfigured_image(self, device_type, version='latest', raw=False):
+        """
+        Download an unconfigured OS image.
+
+        Args:
+            device_type (str): device type slug.
+            version (str): the Balena OS version of the image (defaults to 'latest'). The SDK will try to parse version into semver-compatible version, unsupported (unpublished) version will result in rejection.
```
thgreasi: There is one more
```suggestion
            version (str): the balenaOS version of the image (defaults to 'latest'). The SDK will try to parse version into semver-compatible version, unsupported (unpublished) version will result in rejection.
```

balena/models/device_os.py
```diff
@@ -68,14 +68,21 @@ def get_config(self, app_id, options):
...
         )
 
-    def download(self, raw=None, **data):
+    def download(self, raw=False, **data):
         """
         Download an OS image. This function only works if you log in using credentials or Auth Token.
 
         Args:
             raw (bool): determining function return value.
+            is_dev_image (Optional[bool]): select development image instead of production image. defaults to False.
```
thgreasi: Reminder to also delete the docs :)

balena/models/device_os.py
```diff
@@ -68,14 +68,21 @@ def get_config(self, app_id, options):
             endpoint=self.settings.get('api_endpoint')
         )
 
-    def download(self, raw=None, **data):
+    def download(self, raw=False, **data):
```
thgreasi, line 71: Optional-if-you-agree: It would be nice to default to `latest` in here as well, just for the shake of consistency.
If the required methods for getting the app's device type are not on hand, then that's fine for later.
Alternatively we could drop `'latest'` from `download_unconfigured_image` just to make these consistent and avoid user confusion/errors.

thgreasi (approved):
LGTM :+1: Just 2 docs changes that need addressing before merging :+1:

PR #178, opened 2020-03-12
Fix release tag tests

nghiant2710:
- Fix tag.get_all() test: remove the check for empty tag list by default since user should be able to see tag from public application (ESR apps).

- Update tag.set() and tag.remove() test: to make sure these tests work fine when returned result from tag.get_all() contains release_tag from public apps


nghiant2710:
@sradevski, @thgreasi can I have a review please?

sradevski (approved):

PR #189, opened 2020-06-25
**BREAKING** Add support for the v6 organizations model

nghiant2710:
Connects-to: https://github.com/balena-io/balena-sdk/pull/621
Depends-on: https://github.com/balena-io/balena-api/pull/2018

I'm so sorry for the huge PR, I tried to split them into smaller ones but bulldozer just merged them right away.


balena/models/application.py
```diff
@@ -671,13 +704,23 @@ def set_to_release(self, app_id, full_release_hash):
...
 
+        raw_query = "$filter=startswith(commit, '{release_hash}')&$top=1&select=id&filter=belongs_to__application%20eq%20'{app_id}'%20and%20status%20eq%20'success'".format(
+            release_hash=full_release_hash,
+            app_id=app_id
+        )
+
+        try:
+            release = self.release._Release__get_by_raw_query(raw_query)[0]
+        except exceptions.ReleaseNotFound:
+            raise exceptions.ReleaseNotFound(app_id)
```
thgreasi: I think this should be
```suggestion
            raise exceptions.ReleaseNotFound(full_release_hash)
```

thgreasi (commented):
Just a NIT, otherwise LGTM :+1:

balena/models/application.py
nghiant2710: @thgreasi, thanks for spotting this out. Updated!

thgreasi (approved):

Issue #190, opened 2020-09-03
Should expose the device metadata


vipulgupta2048:
HQ: balena-io/balena-io#2264

thgreasi:
@vipulgupta2048 I think you can close this now :)

PR #191, opened 2020-09-03
Implement device.get_metrics()

vipulgupta2048:
Change-type: minor
Solves #190 

Todo:
- Get the documentation script working with python3.8
- Get some tests done

### DIFF
--- a/balena/models/device.py
+++ b/balena/models/device.py
@@ -1592,3 +1592,25 @@ class Device(object):
             return device['mac_address'].split()
         else:
             return []
+
+    def get_metrics(self, uuid):
+        """
+        Get the metrics related information for a device.
+
+        Args:
+            uuid (str): device uuid.
+
+        Returns:
+            dict: metrics of the device.
+
+        Raises:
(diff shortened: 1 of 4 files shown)


balena/models/device.py
```diff
@@ -1592,3 +1592,34 @@ def get_mac_address(self, uuid):
...
+
+        Raises:
+            DeviceNotFound: if device couldn't be found.
+
+        """
+
+        device = self.get(uuid)
+        if "memory_usage" in device:
+            return {
+                "memory_usage": device["memory_usage"],
```
nghiant2710: can we make a list of all fields we want and extract it like this?
```
metrics = ['memory_usage', 'memory_total']
return { k: device.get(k, '') for k in metrics }
```

It will be a bit nicer and we don't have to repeat all the fields

CameronDiver (dismissed):

CameronDiver (approved):

thgreasi:
HQ: https://github.com/balena-io/balena-io/issues/2264

PR #193, opened 2020-09-19
HOME_DIRECTORY to /tmp

tekollt:
To work with AWS Lambda Functions, a writable location has to be used as HOME_DIRECTORY

### DIFF
--- a/balena/settings.py
+++ b/balena/settings.py
@@ -26,7 +26,7 @@ class Settings(object):
 
     """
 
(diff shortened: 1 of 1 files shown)


nghiant2710:
@tekollt, thank you for the PR, I'm sorry but we can't just change the HOME_DIRECTORY like that as It will be a breaking change. I think it's better if we can add a check to see the default HOME_DIRECTORY is writable or not and use `/tmp` if it's not writable.

You can create an issue and I can quickly implement it :)

PR #198, opened 2020-12-04
Support fetching devices by short UUID

nbertram:
This PR switches `models.device.get()` to filter using `startswith()` instead of `eq` so the short 7-char UUIDs can be resolved. This brings it into line with the JS SDK.

Solves https://github.com/balena-io/balena-sdk-python/issues/197

### DIFF
--- a/balena/models/device.py
+++ b/balena/models/device.py
@@ -83,13 +83,10 @@ class Device(object):
 
         """
 
-        params = {
-            'filter': 'uuid',
-            'eq': uuid
-        }
+        raw_query = "$filter=startswith(uuid,'{uuid}')".format(uuid=uuid)
         try:
             devices = self.base_request.request(
-                'device', 'GET', params=params,
+                'device', 'GET', raw_query=raw_query,
                 endpoint=self.settings.get('pine_endpoint')
             )['d']
             if len(devices) > 1:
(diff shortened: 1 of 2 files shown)


nghiant2710:
@nbertram, thanks a lot for the PR, LGTM but I think we should put this on hold since this change should be applied for other methods as well.

I will merge your PR when I implement the change for others as well. Hopefully we can have it soon.

nbertram:
Thanks for the review. IMO the query param construction logic probably needs to be enhanced to handle more than 'eq' filters, to avoid all these calls having to drop down to raw_query.

The rationale for the JS SDK to support short UUIDs probably came about because the CLI uses it and needs to take it as user input, it's likely more niche in regular SDK usage. In my case I have users entering the short UUID into a management app so I need it for that.

I'm not sure if there's a performance cost to using startsWith when we don't need to. Openbalena seems to issue a LIKE query against Postgres when using it. I couldn't discern too much of a difference against the Cloud.

Also on that topic, the unit tests in the python SDK don't run neatly against openbalena due to missing app types and public URL etc logic that's missing. Not sure if that's a priority at all to sort out, just thought I'd mention it because that was how I first tried to run the tests. Running the tests against an account in the cloud quickly hit the API ratelimit!

PR #205, opened 2021-04-23
Manage lock overrides using the config variables instead of the endpoint

### LINKED ISSUE #203: Should use the update lock override config vars directly
[thgreasi] Should stop relying on the `/device/${deviceId}/lock-override` endpoint, which is going to get deprecated and removed and set & read the RESIN_OVERRIDE_LOCK directly.

Discussed in arch call 2021-04-06
See: https://jel.ly.fish/32cc2ee4-c5a7-4ec8-8e91-1ba4f1379bc4
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/6nQeDu8vFurrNVfNHMUFr1YqtPi
Depends-on: https://github.com/balena-io/open-balena-api/issues/632
Depends-on: https://github.com/balena-io/balena-api/issues/3076

nghiant2710:
This also allows the device lock override methods to work on open-balena

Resolves: #203 
Connects-to: https://github.com/balena-io/balena-sdk/pull/1072
Change-type: minor

### DIFF
--- a/balena/models/device.py
+++ b/balena/models/device.py
@@ -26,6 +26,7 @@ from .hup import Hup
 LOCAL_MODE_MIN_OS_VERSION = '2.0.0'
 LOCAL_MODE_MIN_SUPERVISOR_VERSION = '4.0.0'
 LOCAL_MODE_ENV_VAR = 'RESIN_SUPERVISOR_LOCAL_MODE'
+OVERRIDE_LOCK_ENV_VAR = 'RESIN_OVERRIDE_LOCK'
 
 
 # TODO: support both device uuid and device id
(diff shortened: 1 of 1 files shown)


nghiant2710:
@thgreasi, can I have a review for this PR please?

balena/models/device.py
```diff
@@ -63,6 +64,58 @@ def __init__(self):
...
+                'name': name,
+                'value': value
+            }
+
+            return self.base_request.request(
+                'device_config_variable', 'POST', data=data,
+                endpoint=self.settings.get('pine_endpoint')
+            )
+        except exceptions.RequestError as e:
+            if 'Unique key constraint violated' in e.message or 'must be unique' in e.message:
```
thgreasi, line 81: Here is how pineClient does this check in upsert.
See: https://github.com/balena-io-modules/pinejs-client-js/blob/4a4f8535ef5b3d824ebbc09b46589efecb3a73fe/src/index.ts#L1079-L1080

We can be sure that this will work since model v6, so the current major of the python SDK should probably be able to use it w/o issues.

thgreasi (approved):

PR #206, opened 2021-04-23
Fix wrong example for DeviceServiceEnvVar method

nghiant2710:
Fixing wrong example for models.environment_variables.device_service_environment_variable.get_all_by_application(). Current example is for device env var not device service env var.

Change-type: patch


nghiant2710:
@balena-ci I self-certify!

PR #207, opened 2021-04-23
Cleanup

nghiant2710:
- Fix dependencies.
- Update error message in test case.


nghiant2710:
@balena-ci I self-certify!

PR #212, opened 2021-07-22
Add SDK methods for org invite

nghiant2710:
Connects-to: https://github.com/balena-io/balena-sdk/pull/1008
Change-type: minor

### DIFF
--- a/balena/models/organization.py
+++ b/balena/models/organization.py
@@ -26,6 +26,7 @@ class Organization (object):
         self.settings = Settings()
         self.config = Config()
         self.auth = Auth()
+        self.invite = OrganizationInvite()
 
     def create(self, name, handle=None):
         """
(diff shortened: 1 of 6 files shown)


balena/models/organization.py
```diff
@@ -161,3 +162,143 @@ def remove(self, org_id):
...
+            endpoint=self.settings.get('pine_endpoint')
+        )['d']
+        
+    def create(self, org_id, invitee, role_name=None, message=None):
+        """
+        Creates a new invite for an organization.
+
+        Args:
+            org_id (str): organization id.
+            invitee (str): the email/balena_username of the invitee.
```
thgreasi, line 229: That's actually something we need to change in the JS SDK as well
```suggestion
            invitee (str): the email of the invitee.
```
See: https://github.com/balena-io/balena-api/blob/a0472380577b2fd516c3d64929fb42af86830b22/src/features/invitations/hooks/application.ts#L74-L76

thgreasi (approved):
Other then the str vs int for the id params (for which I'm no sure what is the correct) LGTM :+1:

PR #215, opened 2021-07-23
Add application membership and invite

nghiant2710:
Connects-to: https://github.com/balena-io/balena-sdk/pull/1050, https://github.com/balena-io/balena-sdk/pull/894
Change-type: minor

### DIFF
--- a/balena/models/application.py
+++ b/balena/models/application.py
@@ -32,6 +32,7 @@ class Application(object):
         self.config = Config()
         self.auth = Auth()
         self.release = Release()
+        self.invite = ApplicationInvite()
 
     def __get_single_install_summary(self, raw_data):
         """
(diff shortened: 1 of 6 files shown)


balena/models/application.py
```diff
@@ -881,3 +914,300 @@ def get_dashboard_url(self, app_id):
...
+            self.RESOURCE, 'GET', params=params,
+            endpoint=self.settings.get('pine_endpoint')
+        )['d']
+        
+    def create(self, app_id, invitee, role_name=None, message=None):
+        """
+        Creates a new invite for an application.
+
+        Args:
+            app_id (str): application id.
```
thgreasi: Shouldn't the `str` here be `int`?

balena/models/application.py
```diff
@@ -881,3 +914,300 @@ def get_dashboard_url(self, app_id):
...
+            invitee (str): the email/balena_username of the invitee.
```
thgreasi: That's actually something we need to change in the JS SDK as well
```suggestion
            invitee (str): the email of the invitee.
```
See: https://github.com/balena-io/balena-api/blob/a0472380577b2fd516c3d64929fb42af86830b22/src/features/invitations/hooks/application.ts#L74-L76

balena/models/application.py
```diff
@@ -881,3 +914,300 @@ def get_dashboard_url(self, app_id):
...
+        
+        data = {
+            'username': user_name,
+            'is_member_of__application': app_id
+        }
+
+        if role_name:
+            data['application_membership_role '] = _get_role_by_name(role_name)
+
+        print(data)
```
thgreasi: `print` :police_car:

thgreasi (commented):
Feel free to self-certify after confirming whether we should be documenting the id params as str or int.

nghiant2710:
@balena-ci I self-certify!

PR #217, opened 2021-07-23
Balena SDK for Python v11

nghiant2710:


balena/models/organization.py
```diff
@@ -161,3 +164,241 @@ def remove(self, org_id):
...
+        """
+
+        params = {
+            'filter': 'organization_membership',
+            'eq': membership_id
+        }
+
+        return super(OrganizationMembershipTag, self).get_all(params=params)
+
+    def get_all(self):
```
thgreasi, line 493: How about also accepting a `params` argument here?
* In general we do allow that in all getAll method in the JS sdk
* If we do so, then all other methods could use the getAll of the current instance, instead of the base class.

balena/twofactor_auth.py
```diff
@@ -131,5 +133,79 @@ def get_otpauth_secret(self):
...
         )
-        tmp = urlparse.parse_qs(otp_auth_url)
-        return tmp['secret'][0]
+        tmp = parse_qs(otp_auth_url)
+        return tmp['secret'.encode()][0]
+
+    def verify(self, code):
+        """
+        Verifies two factor authentication.
+        Note that this method not update the token automatically. You should use balena.twofactor_auth.challenge() when possible, as it takes care of that as well.
```
thgreasi, line 142:
```suggestion
        Note that this method does not update the token automatically. You should use balena.twofactor_auth.challenge() when possible, as it takes care of that as well.
```

thgreasi (commented):
Other than that LGTM :+1:
Feel free to self-certify.

nghiant2710:
@balena-ci test!

nghiant2710:
@balena-ci test

nghiant2710:
@balena-ci I self-certify!

PR #221, opened 2021-11-15
patch: Update balena API url

klutchell:
See https://github.com/balena-io/docs/issues/2121


klutchell:
@balena-ci I self-certify!

PR #223, opened 2021-12-29
Update application.generate_provisioning_key to include key name and description

nghiant2710:
Change-type: minor

### DIFF
--- a/balena/models/application.py
+++ b/balena/models/application.py
@@ -727,12 +727,15 @@ class Application:
             endpoint=self.settings.get('pine_endpoint')
         )
 
-    def generate_provisioning_key(self, app_id):
+    def generate_provisioning_key(self, app_id, key_name=None, description=None, expiry_date=None):
         """
         Generate a device provisioning key for a specific application.
 
         Args:
             app_id (str): application id.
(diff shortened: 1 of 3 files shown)


balena/models/application.py
```diff
@@ -742,9 +744,17 @@ def generate_provisioning_key(self, app_id):
         # Make sure user has access to the app_id
         self.get_by_id(app_id)
 
+        data = {
+            'actorType': 'application',
+            'actorTypeId': app_id,
+            'roles': ['provisioning-api-key'],
+            'name': key_name,
+            'description': description
```
thgreasi: You can also add the `expiryDate` if you want, but I think it should be defaulting to Null rather than empty string.
```suggestion
            'description': description,
            'expiryDate': expiry_date
```

thgreasi (commented):
LGTM feel free to self-certify (even if you prefer to only add the expiry date after the major).

balena/models/application.py
nghiant2710: I added expiry_date as an optional parameter and it defaults to None.

nghiant2710:
@balena-ci I self-certify!

ghost (approved):

PR #224, opened 2021-12-29
Few updates for api_key model

nghiant2710:
This PR contains the following changes:
```
- Add api_key.get_provisioning_api_keys_by_application method 
- Add api_key.get_device_api_keys_by_device method 
- Implement api_key.get_all_named_user_api_keys method
```

### DIFF
--- a/balena/auth.py
+++ b/balena/auth.py
@@ -12,6 +12,7 @@ class Auth:
     """
 
     _user_detail_cache = {}
+    _user_full_detail_cache = {}
 
     def __init__(self):
         self.base_request = BaseRequest()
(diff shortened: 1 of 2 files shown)


balena/auth.py
```diff
@@ -12,6 +12,7 @@ class Auth(object):
     """
 
     _user_detail_cache = {}
+    _user_full_detail_cache = {}
```
thgreasi, line 15: I guess we should also clean this up from within `login`, `login_with_token` & `log_out`, like we do for `_user_detail_cache` , right?

thgreasi (commented):
Other than that LGTM, feel free to self-certify

balena/auth.py
nghiant2710: updated.

nghiant2710:
@balena-ci I self-certify!

PR #226, opened 2022-01-05
application.create: Rely on the hostApps for detecting discontinued DTs

nghiant2710:
Connects-to: https://github.com/balena-io/balena-sdk/pull/1139

Change-type: patch

### DIFF
--- a/balena/models/application.py
+++ b/balena/models/application.py
@@ -385,20 +385,24 @@ class Application(object):
         device_manifest = [device for device in device_types if device['name'] == device_type]
 
         if device_manifest:
-            if device_manifest[0]['state'] == 'DISCONTINUED':
-                raise exceptions.BalenaDiscontinuedDeviceType(device_type)
-
-            params = {
-                'filter': 'slug',
-                'eq': device_manifest[0]['slug']
-            }
+            raw_query = "$filter=(slug%20eq%20'{slug}')%20or%20(name%20eq%20'{slug}')&$select=id,name&$expand=is_default_for__application($select=is_archived&$filter=is_host%20eq%20true)".format(slug=device_manifest[0]['slug'])
 
             device_type_detail = self.base_request.request(
-                'device_type', 'GET', params=params,
+                'device_type', 'GET', raw_query=raw_query,
                 endpoint=self.settings.get('pine_endpoint'), login=True
             )['d'][0]
+            
+            if not device_type_detail:
+                raise exceptions.InvalidDeviceType(device_type)
+
         else:
             raise exceptions.InvalidDeviceType(device_type)
+
+        host_apps = device_type_detail['is_default_for__application']
+        # TODO: We are now checking whether all returned hostApps are marked as archived so that we
+		# do not break open-balena. Once open-balena gets hostApps, we can change this to just a $filter on is_archived.
+        if (len(host_apps) > 0 and all([dt['is_archived'] for dt in host_apps])):
+            raise exceptions.BalenaDiscontinuedDeviceType(device_type)
 
         data = {
             'app_name': name,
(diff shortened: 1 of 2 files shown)


balena/models/application.py
```diff
@@ -382,24 +382,24 @@ def create(self, name, device_type, organization, app_type=None):
                 raise exceptions.OrganizationNotFound(organization)
 
         device_types = self.config.get_device_types()
-        device_manifest = [device for device in device_types if device['name'] == device_type]
+        device_manifest = [device for device in device_types if (device['name'] == device_type) or (device_type in device.get('aliases', []))]
```
thgreasi: It seems that we already didn't support DT aliases, in which case I would highly suggest to replace `config.get_device_types()` with a query to the `/v6/device_type` resource filtering on the `slug $or name`.

I'm already working on adding the device type aliases to the model, so we can start using them once that is merged.
See: https://github.com/balena-io/open-balena-api/pull/898

To be clear that's part of the effort of reducing our dependence and dropping the deprecated `/device-types/v1` endpoint.

```suggestion
```

balena/models/application.py
```diff
@@ -382,24 +382,24 @@ def create(self, name, device_type, organization, app_type=None):
...
 
         if device_manifest:
-            if device_manifest[0]['state'] == 'DISCONTINUED':
-                raise exceptions.BalenaDiscontinuedDeviceType(device_type)
-
-            params = {
-                'filter': 'slug',
-                'eq': device_manifest[0]['slug']
-            }
+            raw_query = "$filter=slug%20eq%20'{slug}'&$select=id,name&$expand=is_default_for__application($select=is_archived&$filter=is_host%20eq%20true)".format(slug=device_manifest[0]['slug'])
```
thgreasi:
```suggestion
            raw_query = "$filter=(slug%20eq%20'{slug}')%20or%20(name%20eq%20'{slug}')&$select=id,name&$expand=is_default_for__application($select=is_archived&$filter=is_host%20eq%20true)".format(slug=device_manifest[0]['slug'])
```
and make that request outside of the `if`
and change the `if` to
```
if not device_type_detail:
    raise exceptions.InvalidDeviceType(device_type)
```

balena/models/application.py
nghiant2710: @thgreasi, I know we're changing to the new endpoint but since the latest SDK uses `/device-types/v1` (https://github.com/balena-io/balena-sdk/blob/master/lib/models/config.ts#L158) so I still want to keep using it here and will make a change when the JS SDK updated.

balena/models/application.py
nghiant2710: Updated, the request should be inside the `if ` since there will be exception if `device_manifest` is empty.

balena/models/application.py
thgreasi: @nghiant2710 the only reason that the JS SDK still uses it is in order to support aliases.
Given that the python SDK doesn't support aliases, I don't see any reason use it just for the shake of using it.
I will leave this up to you though, since we can do this change in a follow up.

balena/models/application.py
thgreasi: I should have said
```
raw_query = "$filter=(slug%20eq%20'{device_type}')%20or%20(name%20eq%20'{device_type}')&$select=id,name&$expand=is_default_for__application($select=is_archived&$filter=is_host%20eq%20true)".format(device_type=device_type)
```
But w/o removing the `/device-types/v1` check, this doesn't have any effect and the second path of the `or` will never be true.

Feel free to revert to the original version of the query if you end up keeping the slug resolution via the `/device-types/v1` .

thgreasi:
@nghiant2710 feel free to self-certify :+1:

balena/models/application.py
nghiant2710: @thgreasi, I just realise that the output of `/v6/device_type` endpoint contains less information than current `/device-types/v1` which potentially breaks other methods so it's better to do this change in another PR.

nghiant2710:
@balena-ci I self-certify!

ghost (approved):

PR #230, opened 2022-02-08
Fix invalid semver string when parsing balena os version

### LINKED ISSUE #227: ValueError("2.88.4.prod is not valid SemVer string")
We're using version`9.5.0`, I'm assuming this may be resolved in the latest version? But I can't upgrade due to https://github.com/balena-io/balena-sdk-python/pull/219 not being merged. Thanks for your help.

```python
balena.models.device_os.get_supported_versions("raspberrypi4-64")
```

### LINKED ISSUE #244: Should stop using /device-types/v1/:deviceType/images in favor of the hostApp releases model in the python sdk
[thgreasi] undefined

nghiant2710:
Fixes #227 
Fixes #244

Change-type: patch

### DIFF
--- a/balena/models/device_os.py
+++ b/balena/models/device_os.py
@@ -20,6 +20,40 @@ ARCH_COMPATIBILITY_MAP = {
     'armv7hf': ['rpi']
 }
 
+def cmp_to_key(mycmp):
+    'Convert a cmp= function into a key= function'
+    class K:
+        def __init__(self, obj, *args):
+            self.obj = obj
+        def __lt__(self, other):
+            return mycmp(self.obj, other.obj) < 0
+        def __gt__(self, other):
+            return mycmp(self.obj, other.obj) > 0
+        def __eq__(self, other):
+            return mycmp(self.obj, other.obj) == 0
+        def __le__(self, other):
+            return mycmp(self.obj, other.obj) <= 0
+        def __ge__(self, other):
+            return mycmp(self.obj, other.obj) >= 0
+        def __ne__(self, other):
+            return mycmp(self.obj, other.obj) != 0
+    return K
+
+def sort_version(x, y):
+    if semver.VersionInfo.isvalid(x['raw_version']) and semver.VersionInfo.isvalid(y['raw_version']):
+        return semver.compare(x['raw_version'], y['raw_version'])
+    else:
+        # return 0 if they are not valid semver
+        return 0
+
+def bsemver_match_range(version, version_range):
+    if semver.VersionInfo.isvalid(version):
+        try:
+            if version_range and semver.match(version, ">{version_range}".format(version_range=version_range)):
+                return True
+        except:
+            return False
+    return False
 
 # TODO: https://github.com/balena-io/balena-sdk/pull/288
 class DeviceOs:
@@ -29,6 +63,22 @@ class DeviceOs:
     """
 
     OS_UPDATE_ACTION_NAME = 'resinhup'
+    RELEASE_POLICY_TAG_NAME = 'release-policy'
+    ESR_NEXT_TAG_NAME = 'esr-next'
+    ESR_CURRENT_TAG_NAME = 'esr-current'
+    ESR_SUNSET_TAG_NAME = 'esr-sunset'
+    VARIANT_TAG_NAME = 'variant'
+    VERSION_TAG_NAME = 'version'
+    BASED_ON_VERSION_TAG_NAME = 'meta-balena-base'
+    OS_TYPES = {
+        'default': 'default',
+        'esr': 'esr'
+    }
+    OS_VARIANTS = {
+        'production': 'prod',
+        'development': 'dev'
+    }
+    
 
     def __init__(self):
         self.base_request = BaseRequest()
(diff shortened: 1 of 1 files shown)


nghiant2710:
@thgreasi, can I have a review for this PR please?

balena/models/device_os.py
```diff
@@ -274,14 +274,24 @@ def get_supported_versions(self, device_type):
...
 
         response = self.base_request.request(
             '/device-types/v1/{device_type}/images'.format(device_type=device_type), 'GET',
             endpoint=self.settings.get('api_endpoint'), auth=False
         )
 
         potential_recommended_versions = [i for i in response['versions'] if not re.search(r'(\.|\+|-)dev', i)]
-        potential_recommended_versions = [i for i in potential_recommended_versions if not semver.parse(i)['prerelease']]
+        
+        # returned versions contain '.prod' suffix (like 2.88.4.prod or 2.87.16+rev1.prod) which is not a valid semver string
```
thgreasi: @nghiant2710 `2.87.16+rev1.prod` is a valid semver. `2.88.4.prod` isn't but we shouldn't have such release for any DT. I think we did have this specific invalid version for 4 DTs but we renamed them to `2.88.4+rev0.prod`. Do you have an example for reproducing this?

balena/models/device_os.py
```diff
@@ -274,14 +274,24 @@ def get_supported_versions(self, device_type):
                 * default - recommended (if available) or latest otherwise.
 
         """
+        
+        prod_suffix = '.prod'
 
         response = self.base_request.request(
             '/device-types/v1/{device_type}/images'.format(device_type=device_type), 'GET',
```
thgreasi: Reminder that we also need to change this to be using the device_type resource.

balena/models/device_os.py
```diff
@@ -274,14 +274,24 @@ def get_supported_versions(self, device_type):
...
         )
 
         potential_recommended_versions = [i for i in response['versions'] if not re.search(r'(\.|\+|-)dev', i)]
-        potential_recommended_versions = [i for i in potential_recommended_versions if not semver.parse(i)['prerelease']]
+        
+        # returned versions contain '.prod' suffix (like 2.88.4.prod or 2.87.16+rev1.prod) which is not a valid semver string
+        # we need to remove the suffix before passing to semver.parse()
+        potential_recommended_versions = [
+            i for i in potential_recommended_versions if not semver.parse(
+                i[:-len(prod_suffix)] if i.endswith(prod_suffix) else i
```
thgreasi: In balena-semver we would be checking whether there is any `+` before` the `.prod/.dev` suffix, and iff not then change the `.` to a `+` so that's parse-able. But on the other hand we only do that for sorting purposes and the version strings in the `potential_recommended_versions` should be the original ones, not the normalized ones, otherwise the img-maker will not be able to find a matching image on S3.

balena/models/device_os.py
thgreasi: Oh I see that it's for `raspberrypi4-64` for `https://api.balena-cloud.com/device-types/v1/raspberrypi4-64/images`...
The `2.88.4.prod` seems like a leftover from the rename (right @alexgg ?), except if we intentionally left it there just in case anyone uses it, which they shouldn't.
I think the proper fix for this is to switch from `/device-types/v1/{device_type}/images` to the `/v6/device_type` endpoint.

nghiant2710: hey @thgreasi, with the new endpoint `/v6/device_type`, how do I select all the supported OS version for a device type?

thgreasi: For future reference, this was answered in FD:
See: https://www.flowdock.com/app/rulemotion/platform/threads/_CF4HyplNIvKi_SH2EHHqJKP1WC

nghiant2710:
@thgreasi, finally switched to hostApps, can I have a review for this PR again?
With this and #236, we no longer have any references of `/device-types/v1/` endpoint in Python SDK.

balena/models/device_os.py
```diff
@@ -259,37 +310,145 @@ def get_device_os_semver_with_variant(self, os_version, os_variant=None):
...
+        for device_type in os_versions_by_device_type:
+            os_versions_by_device_type[device_type].sort(reverse=True, key=cmp_to_key(sort_version))
+            recommended_per_os_type = {}
+            
+            for version in os_versions_by_device_type[device_type]:
+                if version['os_type'] not in recommended_per_os_type:
+                    if version['variant'] != 'dev' and not version['known_issue_list'] and not version['version']['prerelease']:
+                        additional_format = ' ({line}, recommended)'.format(line=version['line']) if version['line'] else ' (recommended)'
+                        version['is_recommended'] = True
+                        version['formatted_version'] = 'v{version}{additional_format}'.format(version=version['stripped_version'], additional_format=additional_format)
```
thgreasi, line 430: Given that those two are already deprecated in the node sdk, I wouldn't mind if you never added them at all in the python sdk ;)

balena/models/device_os.py
```diff
@@ -1,3 +1,4 @@
+from audioop import reverse
```
thgreasi: Is this really the correct import?
> The [audioop](https://docs.python.org/3/library/audioop.html#module-audioop) module contains some useful operations on sound fragments.

balena/models/device_os.py
```diff
@@ -259,37 +310,145 @@ def get_device_os_semver_with_variant(self, os_version, os_variant=None):
...
+        releases_with_os_versions = []
+
+        for release in releases:
+            tag_map = {}
+            
+            for release_tag in release['release_tag']:
+                tag_map[release_tag['tag_key']] = release_tag['value']
+
+            if release['raw_version'].startswith('0.0.0'):
+                # TODO: Drop this `else` once we migrate all version & variant tags to release.semver field
```
thgreasi, line 361: This ended up being the `if` part :)

balena/models/device_os.py
```diff
@@ -259,37 +310,145 @@ def get_device_os_semver_with_variant(self, os_version, os_variant=None):
...
+                        version=version,
+                        non_variant_builds='.'.join(non_variant_build_parts)
+                    )
+
+                variant = next((variant for variant in release['version']['build'] if variant in os_variant_keywords) , None)
+                # Backfill the native rel
+				# TODO: This potentially generates an invalid semver and we should be doing
+				# something like `.join(!version.includes('+') ? '+' : '.')`,  but this needs
+				# discussion since otherwise it will break all ESR released as of writing this.
+                release['raw_version'] = '.'.join([x for x in [version, variant] if x])
```
thgreasi: This backfill should be at the end of the `if release['raw_version'].startswith('0.0.0'):` branch.

balena/models/device_os.py
```diff
@@ -259,37 +310,145 @@ def get_device_os_semver_with_variant(self, os_version, os_variant=None):
...
+            based_on_version = tag_map[self.BASED_ON_VERSION_TAG_NAME] if self.BASED_ON_VERSION_TAG_NAME in tag_map else version
+            line = self.__get_os_version_release_line(version, app_tags)
+
+            release.update({
+                'os_type': app_tags['os_type'],
+                'line': line,
+                'stripped_version': version,
+                'based_on_version': based_on_version,
+                'variant': variant,
+                'formatted_version': 'v{version}{line}'.format(version=version,line=f' ({line})' if line else '')
```
thgreasi, line 400: Given that those this is already deprecated in the node sdk, I wouldn't mind if you never added it at all in the python sdk ;)

balena/models/device_os.py
```diff
@@ -259,37 +310,145 @@ def get_device_os_semver_with_variant(self, os_version, os_variant=None):
...
+        return {
+            'os_type': tag_map[self.RELEASE_POLICY_TAG_NAME] if self.RELEASE_POLICY_TAG_NAME in tag_map else self.OS_TYPES['default'],
+            'next_line_version_range': tag_map[self.ESR_NEXT_TAG_NAME] if self.ESR_NEXT_TAG_NAME in tag_map else '',
+            'current_line_version_range': tag_map[self.ESR_CURRENT_TAG_NAME] if self.ESR_CURRENT_TAG_NAME in tag_map else '',
+            'sunset_line_version_range': tag_map[self.ESR_SUNSET_TAG_NAME] if self.ESR_SUNSET_TAG_NAME in tag_map else ''
+        }
+
+    def __get_os_versions(self, device_type):
+        raw_query = "$select=is_for__device_type&$expand=application_tag($select=tag_key,value),is_for__device_type($select=slug)"
+        raw_query += ",owns__release($select=id,raw_version,semver,release_version,version,is_final,release_tag,known_issue_list&$filter=is_final%20eq%20true&$filter=is_invalidated%20eq%20false&$filter=status%20eq%20'success'&$expand=release_tag)"
```
thgreasi: I think we don't use some of the selected props
```suggestion
        raw_query += ",owns__release($select=id,raw_version,version,known_issue_list&$filter=is_final%20eq%20true&$filter=is_invalidated%20eq%20false&$filter=status%20eq%20'success'&$expand=release_tag)"
```

balena/models/device_os.py
```diff
@@ -259,37 +310,145 @@ def get_device_os_semver_with_variant(self, os_version, os_variant=None):
...
-        potential_recommended_versions = [i for i in response['versions'] if not re.search(r'(\.|\+|-)dev', i)]
-        potential_recommended_versions = [i for i in potential_recommended_versions if not semver.parse(i)['prerelease']]
-        recommended = potential_recommended_versions[0] if potential_recommended_versions else None
-
-        return {
-            'versions': response['versions'],
-            'recommended': recommended,
-            'latest': response['latest'],
-            'default': recommended if recommended else response['latest']
-        }
```
thgreasi, line 293: Given that the return type changed, should the be a major?

thgreasi (commented):
:heart: :heart: :heart:

balena/models/device_os.py
nghiant2710: @thgreasi, I still see these properties in the master branch of JS SDK, will they be removed soon?

balena/models/device_os.py
nghiant2710: Removed, I added it by accident.

balena/models/device_os.py
nghiant2710: My bad, updated!

balena/models/device_os.py
nghiant2710: `semver,release_version` removed. I do use `is_final` and `release_tag` in some private methods

balena/models/device_os.py
nghiant2710: Since there is an user blocked by this and we don't completely change the return object so I marked it as `minor` so it can be released soon but this should be a major.

nghiant2710: Changed to `major`. Will try to finish other changes asap so this can be released soon.

balena/models/device_os.py
thgreasi: @nghiant2710 I've already deprecated them in the current version and going to drop them in the next major.
See: https://github.com/balena-io/balena-sdk/blob/cc5a64f364876d6b9d0585c27f2eadc9c936114f/lib/models/os.ts#L73-L76
The reasoning is that if we want to move closer to the model & that the latest version should be the one that's always recommended by our clients anyway.
The CLI & UI was using those props so far but they have now moved away afair, so given that no client depends on them on the python-sdk I was thinking we could completely avoid adding them and then removing them to be in sync with the node sdk (as it will be after the majro bump).

balena/models/device_os.py
thgreasi: WRT `is_final` I don't see it being used atm on master or in this PR :thinking:
See: https://github.com/balena-io/balena-sdk-python/search?q=is_final&type=code
Are you going to start using it in a follow-up PR?

You should be able to drop the `release_tag` from the `$select`, since it's already in the `$expand` list & I don't think `$select`ing reverse navigation resources is valid (though pine might not error).

balena/models/device_os.py
thgreasi: @nghiant2710 given that:
* we no longer do pre-releases
* .prod sorts higher that .dev

maybe you can keep it a non-major if you do something like:
```
        return {
            'versions': versions_by_dt[device_type],
            'recommended': versions_by_dt[device_type][0],
            'latest': versions_by_dt[device_type][0],
            'default': versions_by_dt[device_type][0]
        }
```
or you can preserve the existing `potential_recommended_versions` logic.

thgreasi: @nghiant2710 any updates on this?

nghiant2710: @thgreasi, I updated the return value so this PR can be a minor release, will change it again in next major release.

nghiant2710: can you please review the PR again?

thgreasi (approved):

PR #239, opened 2022-03-17
Require authentication for device_os.get_supported_version

nghiant2710:
So the SDK can query private device types

Change-type: minor


alexgg (approved):
lgtm

I assume the authentication is already cached in settings.

thgreasi:
@nghiant2710 could auth be optional for this method? As in, it passes the auth header if the instance is logged in or doesn't if the instance isn't logged in.

nghiant2710:
@thgreasi, yes we can, I will make another PR for this.

PR #241, opened 2022-03-19
Fix missing arch in device type information

### LINKED ISSUE #240: Move device fails with 'arch' KeyError
Exception when trying to move a device to an application - since 11.2.0.

The error occurs in `/usr/local/lib/python3.6/dist-packages/balena/models/device.py(951)`
```
if not self.device_os.is_architecture_compatible_with(device_dev_type['arch'], app_dev_type['arch']):
```

Because `device_dev_type` has no key 'arch'.

This happens in 11.2.0 and not in 11.1.0 (just confirmed).

In both version I reference the same device in Balena and trying to move to same application.

It happens with both arm and amd devices.

nghiant2710:
Fixes #240 
Arch information now returned as slug in is_of__cpu_architecture

Change-type: patch


nghiant2710:
@balena-ci I self-certify!

PR #242, opened 2022-03-19
Add auth as an optional arg for device_os.get_supported_versions

### LINKED ISSUE #238: get_supported_versions() does not authenticate
As the `get_supported_versions()` does not pass authentication (see https://github.com/balena-io/balena-sdk-python/blob/7fa198426420e1e5255c7e2c711c1cb0f6abd0ba/balena/models/device_os.py#L280), the SDK fails when querying private device types.

nghiant2710:
Fixes #238 

Change-type: patch

### DIFF
--- a/balena/models/device_os.py
+++ b/balena/models/device_os.py
@@ -259,12 +259,13 @@ class DeviceOs:
             xstr(version_info.build) + '.' + os_variant if os_variant and os_variant not in tmp else version_info.build
         )
 
-    def get_supported_versions(self, device_type):
+    def get_supported_versions(self, device_type, auth=True):
         """
         Get OS supported versions.
 
         Args:
-            device_type (str): device type slug
+            device_type (str): device type slug.
+            auth (Optional[bool]): if auth is True then auth header will be added to the request (to get private device types) otherwise no auth header and only public device types returned. Default to True.
 
         Returns:
             dict: the versions information, of the following structure:
@@ -277,7 +278,7 @@ class DeviceOs:
 
         response = self.base_request.request(
             '/device-types/v1/{device_type}/images'.format(device_type=device_type), 'GET',
-            endpoint=self.settings.get('api_endpoint')
+            endpoint=self.settings.get('api_endpoint'), auth=auth
         )
 
         potential_recommended_versions = [i for i in response['versions'] if not re.search(r'(\.|\+|-)dev', i)]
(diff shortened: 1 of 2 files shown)


balena/models/device_os.py
```diff
@@ -277,7 +278,7 @@ def get_supported_versions(self, device_type):
 
         response = self.base_request.request(
             '/device-types/v1/{device_type}/images'.format(device_type=device_type), 'GET',
-            endpoint=self.settings.get('api_endpoint')
+            endpoint=self.settings.get('api_endpoint'), auth=auth
```
thgreasi, line 281: Instead of asking the user to pass true/false w/ an extra argument, could the BaseRequest have an `auto` mode, that checks `self.settings.get(TOKEN_KEY)` and calls `__set_authorization()` in case it holds a key?

thgreasi (commented):
LGTM as is, but I think passing auth if present would be good to have approach for all methods & resources in general, especially since we added public resources.

balena/models/device_os.py
nghiant2710: @thgreasi, the `BaseRequest` class does check for auth token or user api key and automatically add it to the request headers if users don't explicitly set `auth=False` in some methods support that argument so I think it's doing what you mentioned.

I think provide `auth` argument here just allows users to get less data and send fewer requests to the API if they want since not all users need data about private DTs and ignoring authentication means we can skip the check for auth token expired time (sometimes we need to send a request to refresh token) on each base_request request.

thgreasi: @nghiant2710 in that case, can we default auth to None, so that when the user doesn't provide a specific option, BaseRequest does "if logged in then pass the token" trick? That's how the the js sdk does it as well. So it's like the method works depending on the state of the sdk instance the user is calling it on.

To be clear, feel free to self-certify, since that's already covering the use case we cared about.

balena/models/device_os.py
```diff
@@ -259,12 +259,13 @@ def get_device_os_semver_with_variant(self, os_version, os_variant=None):
...
 
-    def get_supported_versions(self, device_type):
+    def get_supported_versions(self, device_type, auth=None):
         """
         Get OS supported versions.
 
         Args:
-            device_type (str): device type slug
+            device_type (str): device type slug.
+            auth (Optional[bool]): if auth is True then auth header will be added to the request (to get private device types) otherwise no auth header and only public device types returned. Default to None.
```
thgreasi: Should it now say that default/None would use the bearer token if the sdk instance is logged in?

thgreasi (commented):
LGTM :+1:

nghiant2710:
@balena-ci I self-certify!

PR #252, opened 2022-07-01
device: Fix error when parsing balenaOS version before updating OS

### LINKED ISSUE #250: Update HostOS receiving undocumented failure mode
When attempting to update a device HostOS with the following command

balena.models.device.start_os_update("some_uuid", "2.95.3+rev1")

I am receiving the following error: 
TypeError: __init__() got an unexpected keyword argument 'release_tag'

I have also verified that the device is on a lower version of HostOS, and am attempting to migrate it to a valid target

Note, I believe the semver is correct, as before I was receiving 
"Invalid target balenaOS version"

When I tried to use "v2.95.3+rev1"

Fleet Type - Raspberry Pi 3

nghiant2710:
Fixes #250 

There is a change in output of device_os.get_supported_versions method
so device.start_os_update fails to parse current balena OS version
correctly.

Change-type: minor

### DIFF
--- a/balena/models/device.py
+++ b/balena/models/device.py
@@ -13,7 +13,7 @@ except ImportError:  # Python 2 imports
 
 from ..base_request import BaseRequest
 from .config import Config
-from .device_os import DeviceOs
+from .device_os import DeviceOs, normalize_balena_semver
 from .device_type import DeviceType
 from ..settings import Settings
 from ..auth import Auth
@@ -1401,16 +1401,16 @@ class Device:
             raise exceptions.OsUpdateError('The uuid of the device is not available')
 
         if 'is_online' not in device_info or not device_info['is_online']:
-            raise exceptions.OsUpdateError('The device is offline: {uuid}'.format(uuid=uuid))
+            raise exceptions.OsUpdateError('The device is offline: {uuid}'.format(uuid=device_info['uuid']))
 
         if 'os_version' not in device_info or not device_info['os_version']:
-            raise exceptions.OsUpdateError('The current os version of the device is not available: {uuid}'.format(uuid=uuid))
+            raise exceptions.OsUpdateError('The current os version of the device is not available: {uuid}'.format(uuid=device_info['uuid']))
 
         if 'is_of__device_type' not in device_info or not device_info['is_of__device_type']:
-            raise exceptions.OsUpdateError('The device type of the device is not available: {uuid}'.format(uuid=uuid))
+            raise exceptions.OsUpdateError('The device type of the device is not available: {uuid}'.format(uuid=device_info['uuid']))
 
         if 'os_variant' not in device_info:
-            raise exceptions.OsUpdateError('The os variant of the device is not available: {uuid}'.format(uuid=uuid))
+            raise exceptions.OsUpdateError('The os variant of the device is not available: {uuid}'.format(uuid=device_info['uuid']))
 
         current_os_version = self.device_os.get_device_os_semver_with_variant(device_info['os_version'], device_info['os_variant'])
         self.hup.get_hup_action_type(device_info['is_of__device_type'][0]['slug'], current_os_version, target_os_version)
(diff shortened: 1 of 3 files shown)


balena/models/device.py
```diff
@@ -1443,7 +1443,7 @@ def start_os_update(self, uuid, target_os_version):
         self.__check_os_update_target(device, target_os_version)
 
         all_versions = self.device_os.get_supported_versions(device['is_of__device_type'][0]['slug'])['versions']
-        if not [v for v in all_versions if semver.compare(target_os_version, v) == 0]:
+        if not [v for v in all_versions if target_os_version == v['stripped_version']]:
```
thgreasi: Afaict this would prevent users to use the full version `2.29.2+rev1.prod`, which sounds like a breaking change.
Afaik the HUP script does expect the fully qualified version, including the `.prod`/`.dev` if the version is a non-unified one.
How about v ?
```suggestion
        if not [v for v in all_versions if target_os_version == v['raw_version']]:
```

balena/models/device_os.py
```diff
@@ -40,8 +40,8 @@ def __ne__(self, other):
     return K
 
 def sort_version(x, y):
-    if semver.VersionInfo.isvalid(x['raw_version']) and semver.VersionInfo.isvalid(y['raw_version']):
-        return semver.compare(x['raw_version'], y['raw_version'])
+    if semver.VersionInfo.isvalid(x['stripped_version']) and semver.VersionInfo.isvalid(y['stripped_version']):
+        return semver.compare(x['stripped_version'], y['stripped_version'])
```
thgreasi: Is that just to avoid the `1.2.3.prod` buggy case?
PS: Reminder that the node balena-semver also takes revisions into consideration while sorting, while the upstream `semver` packages (node & python) simply ignore all build metadata parts in `.compare`.

balena/models/device_os.py
```diff
@@ -419,7 +430,7 @@ def __transform_host_apps(self, host_apps):
             os_versions_by_device_type[host_app_device_type].extend(self.__get_os_versions_from_releases(host_app['owns__release'], app_tags))
 
         for device_type in os_versions_by_device_type:
-            os_versions_by_device_type[device_type].sort(reverse=True, key=cmp_to_key(sort_version))
+            os_versions_by_device_type[device_type].sort(reverse=True, key=lambda x: x['formatted_version'])
```
thgreasi: Is this making the sorting lexicographical (thus 2.89.0 would show before 2.9.0)?

balena/models/device_os.py
nghiant2710: I've implemented a `compare` function based on `balena-semver.compare` so all build metadata parts will not be ignored.

balena/models/device_os.py
nghiant2710: I've implemented a `compare` function based on `balena-semver.compare` and it is used to sort the versions here.

balena/models/device.py
nghiant2710: @thgreasi, I used `stripped_version` here since we do not include the `.prod`/`.dev` build metadata part in the balenaOS version list in the dashboard but you are right, `raw_version` will cover the fully qualified version case. Changed to `raw_version`

nghiant2710:
Hey @thgreasi, can you have your review for this PR again please?

balena/models/device.py
```diff
@@ -1424,8 +1424,8 @@ def start_os_update(self, uuid, target_os_version):
             InvalidParameter|OsUpdateError: if target_os_version is invalid.
 
         Examples:
-            >>> balena.models.device.start_os_update('b6070f4fea5edf808b576123157fe5ec', '2.29.2+rev1.prod')
-            {u'status': u'in_progress', u'action': u'resinhup', u'parameters': {u'target_version': u'2.29.2+rev1.prod'}, u'last_run': 1554490809219L}
+            >>> balena.models.device.start_os_update('b6070f4fea5edf808b576123157fe5ec', '2.29.2+rev1')
```
thgreasi: Let's add the variant back to the example so that's more obvious that we expect the fully qualified OS version, like:
```suggestion
            >>> balena.models.device.start_os_update('b6070f4fea5edf808b576123157fe5ec', '2.29.2+rev1.prod')
            >>> # or for unified OS releases
            >>> balena.models.device.start_os_update('b6070f4fea5edf808b576123157fe5ec', '2.89.0+rev1')
```

balena/models/device_os.py
```diff
@@ -407,16 +491,17 @@ def __transform_host_apps(self, host_apps):
         os_versions_by_device_type = {}
 
         for host_app in host_apps:
-            host_app_device_type = host_app['is_for__device_type'][0]['slug'] if len(host_app['is_for__device_type']) > 0 else None
+            if host_app['is_host']:
```
thgreasi, line 498: `__get_os_versions` already has a $filter with `is_host: true`, so I think we shouldn't need this, right?

thgreasi (commented):
Just NITs, LGTM, feel free to self-certify :clap:

nghiant2710:
@balena-ci I self-certify!

PR #253, opened 2022-07-08
Add name and description field to device.generate_device_key method

nghiant2710:
Change-type: minor

### DIFF
--- a/balena/models/device.py
+++ b/balena/models/device.py
@@ -1070,12 +1070,15 @@ class Device:
 
         return self.set_custom_location(uuid, {'latitude': '', 'longitude': ''})
 
-    def generate_device_key(self, uuid):
+    def generate_device_key(self, uuid, key_name=None, key_description=None, expiry_date=None):
         """
         Generate a device key.
 
         Args:
             uuid (str): device uuid.
+            key_name (Optional[str]): device key name.
+            key_description (Optional[str]): description for device key.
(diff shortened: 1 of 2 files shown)


balena/models/device.py
```diff
@@ -1087,9 +1089,14 @@ def generate_device_key(self, uuid):
         """
 
         device_id = self.get(uuid)['id']
+        
+        data = {
+            'name': key_name,
+            'description': key_description
```
thgreasi: You can also add the expiryDate if you want, but I think it should be defaulting to Null rather than empty string.
```suggestion
            'description': key_description,
            'expiryDate': expiry_date
```

balena/models/device.py
```diff
@@ -1070,12 +1070,14 @@ def unset_custom_location(self, uuid):
 
         return self.set_custom_location(uuid, {'latitude': '', 'longitude': ''})
 
-    def generate_device_key(self, uuid):
+    def generate_device_key(self, uuid, key_name='', key_description=''):
```
thgreasi: Could we default these to Null?
```suggestion
    def generate_device_key(self, uuid, key_name=None, key_description=None):
```

thgreasi (commented):
LGTM feel free to self-certify (even if you prefer to only add the expiry date after the major).

nghiant2710:
@balena-ci I self-certify!

PR #264, opened 2022-07-12
supervisor: Make app_id optional parameter when calling supervisor.ping from outside balena device

### LINKED ISSUE #262: balena.models.supervisor.ping does not requiere an app_id
The signature of balena.models.supervisor.ping accepts a device uuid and an application id. 

https://github.com/balena-io/balena-sdk-python/blob/18790c1635d99122111c51b5cc3668881db161d2/balena/models/supervisor.py#L112

The application id is not needed, but a call will fail if not provided.

https://github.com/balena-io/balena-sdk-python/blob/18790c1635d99122111c51b5cc3668881db161d2/balena/models/supervisor.py#L134

The application id can be removed from the signature.

nghiant2710:
app_id can be read from device information so only device_uuid is needed.
Fix #262, the `app_id` parameter will be removed in next major release.

Change-type: patch

### DIFF
--- a/balena/models/supervisor.py
+++ b/balena/models/supervisor.py
@@ -115,7 +115,7 @@ class Supervisor:
         No need to set device uuid and app_id if command is sent to the API on device.
 
         Args:
-            device_uuid (Optional[str]): device uuid.
+            device_uuid (Optional[str]): device uuid, mandatory if not calling this method from the service container on balena device.
             app_id (Optional[str]): application id.
 
         Returns:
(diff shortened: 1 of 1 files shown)


nghiant2710:
Hey @andreasdotzler, thank you for the report and PR. We cannot drop the `app_id` immediately since it is a breaking change and will be released in next major release. So in this PR, I drop the check for `api_id` when not calling it from a balena device and you no longer need to specify it. The `device_uuid` is enough to get all the neccessary information to call `supervisor/ping` API endpoint.

nghiant2710:
@balena-ci rebase

nghiant2710:
@balena-ci I self-certify!

ghost (approved):

andreasdotzler:
@nghiant2710 thank you for the fast fix

PR #273, opened 2022-11-19
Add release methods

nghiant2710:


nghiant2710:
hey @thgreasi, can I have a review for this PR please?

nghiant2710:
@balena-ci I self-certify!

ghost (approved):

nghiant2710:
@balena-ci I self-certify!

balena/models/release.py
```diff
@@ -293,3 +293,131 @@ def finalize(self, commit_or_id):
             'release', 'PATCH', params=params, data=data,
             endpoint=self.settings.get('pine_endpoint')
         )
+
+    def set_is_invalidated(self, commit_or_id, is_invalidated=False):
```
thgreasi: I think that this should not have a default value so that the user is explicit what they want to do, given this method can change the target release of the fleet.
```suggestion
    def set_is_invalidated(self, commit_or_id, is_invalidated):
```
Or if you still prefer to have a default value, then it might better be True, since it sounds weird that calling set_is_invalidated w/o an explicit value would mark the release as valid.

balena/models/release.py
```diff
@@ -293,3 +293,131 @@ def finalize(self, commit_or_id):
...
+        """
+        Set the is_invalidated property of a release to True or False.
+
+        Args:
+            commit_or_id (str): release commit (str) or id (int).
+            is_invalidated (bool): True for invalidated, False for validated. Default to False
```
thgreasi: Reminder to update depending what you decide about the above comment.
```suggestion
            is_invalidated (bool): True for invalidated, False for validated.
```

balena/models/release.py
```diff
@@ -293,3 +293,131 @@ def finalize(self, commit_or_id):
...
+        data = {
+            'known_issue_list': known_issue_list
+        }
+
+        return self.base_request.request(
+            'release', 'PATCH', params=params, data=data,
+            endpoint=self.settings.get('pine_endpoint')
+        )
+
+    def set_release_version(self, commit_or_id, release_version):
```
thgreasi: How about using `semver` which is the name of the field for consistency?
Especially since `version` is a different field in the model.

balena/models/release.py
```diff
@@ -293,3 +293,131 @@ def finalize(self, commit_or_id):
...
+        """
+        Set a version for a release.
+
+        Args:
+            commit_or_id (str): release commit (str) or id (int).
+            release_version (str): the release version, only support setting the 3-digit semver values.
```
thgreasi: It can actually be any semver valid value now :)
```suggestion
            release_version (str): the release version, only supports semver compliant values.
```

thgreasi (commented):
Other than those NITs, LGTM

balena/models/release.py
nghiant2710: yeah, makes sense. I removed the default value.

nghiant2710:
@thgreasi, I've fixed your comments, can you please have a look again?

balena/models/release.py
```diff
@@ -293,3 +293,131 @@ def finalize(self, commit_or_id):
...
+        data = {
+            'known_issue_list': known_issue_list
+        }
+
+        return self.base_request.request(
+            'release', 'PATCH', params=params, data=data,
+            endpoint=self.settings.get('pine_endpoint')
+        )
+
+    def set_release_version(self, commit_or_id, semver):
```
thgreasi, line 393: @nghiant2710 sorry if I wasn't clear :disappointed:, I meant the method name:
```suggestion
    def set_release_semver(self, commit_or_id, semver):
```

nghiant2710: @thgreasi, it's called `release version` in the dashboard so I'm not sure if we should change the method name to `set_release_semver` as it should match with our dashboard
image

thgreasi: Fair enough

thgreasi (approved):

nghiant2710:
@balena-ci I retest

nghiant2710:
@balena-ci test

nghiant2710:
@balena-ci I self-certify!

PR #280, opened 2023-02-16
Support ESR version in device.start_os_update()

nghiant2710:
Change-type: minor


nghiant2710:
@balena-ci I self-certify!

PR #286, opened 2023-04-06
Avoid device id lookups where possible by querying with uuid directly

Page-:
Change-type: patch

### DIFF
--- a/balena/models/config_variable.py
+++ b/balena/models/config_variable.py
@@ -49,15 +49,10 @@ class DeviceConfigVariable:
 
         """
 
-        device = self.device.get(uuid)
-
-        params = {
-            'filter': 'device',
-            'eq': device['id']
-        }
+        raw_query = "$filter=device/any(d:d/uuid%20eq%20'{uuid}')".format(uuid=uuid)
 
         return self.base_request.request(
-            'device_config_variable', 'GET', params=params,
+            'device_config_variable', 'GET', raw_query=raw_query,
             endpoint=self.settings.get('pine_endpoint')
         )['d']
 
--- a/balena/models/environment_variables.py
+++ b/balena/models/environment_variables.py
@@ -64,13 +64,9 @@ class DeviceEnvVariable:
 
         """
 
-        device = self.device.get(uuid)
-        params = {
-            'filter': 'device',
-            'eq': device['id']
-        }
+        raw_query = "$filter=device/any(d:d/uuid%20eq%20'{uuid}')".format(uuid=uuid)
         return self.base_request.request(
-            'device_environment_variable', 'GET', params=params,
+            'device_environment_variable', 'GET', raw_query=raw_query,
             endpoint=self.settings.get('pine_endpoint')
         )['d']
 
@@ -209,10 +205,7 @@ class DeviceServiceEnvVariable:
 
         """
 
-        # TODO: pine client for python
-        device = self.device.get(uuid)
-
-        query = '$expand=service_install($select=id&$expand=service($select=service_name))&$filter=service_install/any(d:d/device%20eq%20{device_id})'.format(device_id=device['id'])
+        query = "$expand=service_install($select=id&$expand=service($select=service_name))&$filter=service_install/any(si:si/device/any(d:d/uuid%20eq%20'{uuid}'))".format(uuid=uuid)
 
         return self.base_request.request(
             'device_service_environment_variable', 'GET', raw_query=query,
--- a/balena/models/tag.py
+++ b/balena/models/tag.py
@@ -148,14 +148,9 @@ class DeviceTag(BaseTag):
 
         """
 
-        device = self.device.get(uuid)
+        raw_query = "$filter=device/any(d:d/uuid%20eq%20'{uuid}')".format(uuid=uuid)
 
-        params = {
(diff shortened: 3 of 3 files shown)


balena/models/environment_variables.py
```diff
@@ -209,10 +208,7 @@ def get_all(self, uuid):
 
         """
 
-        # TODO: pine client for python
-        device = self.device.get(uuid)
-
-        query = '$expand=service_install($select=id&$expand=service($select=service_name))&$filter=service_install/any(d:d/device%20eq%20{device_id})'.format(device_id=device['id'])
+        query = '$expand=service_install($select=id&$expand=service($select=service_name))&$filter=service_install/any(d:d/device/uuid%20eq%20{uuid})'.format(uuid=uuid)
```
otaviojacobi: @Page- Any reason to prefer .format over [python f string](https://peps.python.org/pep-0498/) like
```suggestion
        query = f'$expand=service_install($select=id&$expand=service($select=service_name))&$filter=service_install/any(d:d/device/uuid%20eq%20{uuid})'
```

balena/models/config_variable.py
```diff
@@ -49,11 +49,9 @@ def get_all(self, uuid):
 
         """
 
-        device = self.device.get(uuid)
-
         params = {
-            'filter': 'device',
-            'eq': device['id']
+            'filter': 'device/uuid',
+            'eq': uuid
```
thgreasi: Would it make a real difference if we changed it to use:
```
raw_query = '$filter=device/any(d:d/uuid%20eq%20{uuid})'.format(uuid=uuid)
```
?

thgreasi (commented):
Other than the catch of these methods not throwing DeviceNotFound errors (which the `device.get()` was doing, LGTM.

balena/models/environment_variables.py
Page-: I was just modifying existing code (and have essentially non-existent python experience) but I'd be happy in general for the python SDK to use whatever is the best-practice/most idiomatic python features - I just have no idea what those would be

balena/models/config_variable.py
fisehara: I think using raw query strings if not necessary makes it harder to maintain, no?
It's the same pattern of using pine client in node rather than writing the URLs by hand.

thgreasi: @fisehara I don't think the python SDK's "lite pine client" implementation supports using a $any, and that's why I suggested to use raw_query. The existing code already does that in all cases that a $any was necessary, and a proper python pine client is a longstanding TODO.

balena/models/tag.py
```diff
@@ -148,14 +148,9 @@ def get_all_by_device(self, uuid):
...
-        device = self.device.get(uuid)
-
-        params = {
-            'filter': 'device',
-            'eq': device['id']
-        }
+        raw_query = '$filter=device/any(d:d/uuid%20eq%20{uuid})'.format(uuid=uuid)
 
-        return super(DeviceTag, self).get_all(params=params)
+        return super(DeviceTag, self).get_all(raw_query=raw_query)
```
otaviojacobi, line 153: nice!

balena/models/tag.py
```diff
@@ -148,14 +148,9 @@ def get_all_by_device(self, uuid):
 
         """
 
-        device = self.device.get(uuid)
-
-        params = {
-            'filter': 'device',
-            'eq': device['id']
-        }
+        raw_query = '$filter=device/any(d:d/uuid%20eq%20{uuid})'.format(uuid=uuid)
```
otaviojacobi: @Page- isnt it missing a quote? e.g.:
```suggestion
raw_query = "$filter=device/any(d:d/uuid%20eq%20'{uuid}')".format(uuid=uuid)
```

otaviojacobi: Yup: When I execute this code it throws: 
```
  File "/home/otavio/git/balena-sdk-python/test.py", line 6, in <module>
    print(b.models.tag.device.get_all_by_device('01ee734daed728a63186229c33ed606b'))
  File "/home/otavio/git/balena-sdk-python/balena/models/tag.py", line 153, in get_all_by_device
    return super(DeviceTag, self).get_all(raw_query=raw_query)
  File "/home/otavio/git/balena-sdk-python/balena/models/tag.py", line 35, in get_all
  File "/home/otavio/git/balena-sdk-python/balena/base_request.py", line 197, in request
    raise exceptions.RequestError(response._content)
balena.exceptions.RequestError: b'"Malformed url: \'/v6/device_tag?$filter=device/any(d:d/uuid%20eq%2001ee734daed728a63186229c33ed606b)\'"'
```

But my suggestion fixes it

otaviojacobi (changes requested):

balena/models/tag.py
thgreasi: @otaviojacobi what python version are you using? B/c this seems to be passing CI tests w/ 3.10.

thgreasi: Oh, I totally see that we do indeed need the quotes... I guess we are missing a test case for this :disappointed:

thgreasi (approved):

otaviojacobi (approved):

PR #290, opened 2023-04-06
Fixes linting and formatting

otaviojacobi:


poetry.lock
thgreasi, line 1: Will this lock the dependencies for clients as well?
If not, then it might be better to avoid creating a lock (like we do in all our node libraries) so that CI is always testing w/ the dependencies that a fresh installation would get, so that we know if new installations of the SDK ever get broken.

otaviojacobi: hey @thgreasi it is not used for dependencies clients, at least not right now. Right now we use a setup.py file which has a [install_requries defining the dependencies](https://github.com/balena-io/balena-sdk-python/blob/master/setup.py#L77) so my goal adding this poetry.lock was just to ensure CI and local environments match the same version. However, in following PRs I am planning on using poetry to manage the dependencies and build of our package (which would make life much easier) and there yes, I would remove the poetry.lock file.

thgreasi (approved):

PR #294, opened 2023-04-10
Make the tests faster by re-using created resources

thgreasi:
Change-type: patch

### DIFF
--- a/tests/functional/models/test_device.py
+++ b/tests/functional/models/test_device.py
@@ -9,17 +9,21 @@ class TestDevice(unittest.TestCase):
 
     helper = None
     balena = None
+    app = None
+    device = None
 
     @classmethod
     def setUpClass(cls):
         cls.helper = TestHelper()
         cls.balena = cls.helper.balena
+        cls.app = cls.balena.models.application.create('FooBar', 'Raspberry Pi 2', cls.helper.default_organization['id'])
 
-    def tearDown(self):
-        # Wipe all apps after every test case.
-        self.helper.wipe_application()
+    @classmethod
+    def tearDownClass(cls):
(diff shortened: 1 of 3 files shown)


tests/functional/models/test_device.py
```diff
@@ -256,124 +233,112 @@ def test_move(self):
...
         }
 
         # should be able to set the location of a device by uuid.
-        self.balena.models.device.set_custom_location(device['uuid'], location)
-        device = self.balena.models.device.get(device['uuid'])
-        self.assertEqual(device['custom_latitude'], '41.383333')
-        self.assertEqual(device['custom_longitude'], '2.183333')
+        self.balena.models.device.set_custom_location(type(self).device['uuid'], location)
+        device = self.balena.models.device.get(type(self).device['uuid'])
+        self.assertEqual(type(self).device['custom_latitude'], '41.383333')
```
otaviojacobi: hey @thgreasi this test fails because what it is actually doing is updating a property on the device itself. when you do `self.balena.models.device.set_custom_location(type(self).device['uuid'], location)` this changes the custom location, because now `type(self).device)` points to the old variable, this breaks. 

I think the main problem here was when copying type(self).device over device you overwrote the device on line 244. I think that changing in test 20 and 21 the assert to

```suggestion
        self.assertEqual(device['custom_latitude'], '41.383333')
        self.assertEqual(device['custom_longitude'], '2.183333')
```

Should fix it

thgreasi: Right @otaviojacobi ... I guess that was a signal that I was indeed tired XD

thgreasi:
I self-certify!

PR #316, opened 2023-04-25
Fix device history filter arguments

fisehara:
Device history was not filtered by `deviceId`, `deviceUUID` or `applicationID` as SDK created two `$filter` arguments which are not interpreted on the API site.

Change-type: patch

### DIFF
--- a/balena/models/history.py
+++ b/balena/models/history.py
@@ -40,25 +40,26 @@ class HistoryBaseClass(ABC):
             else:
                 raise exceptions.InvalidParameter("toDate", toDate)
 
-        rawFilterString = "$filter=" + " and ".join(rawTimeRangeFilters)
+        return rawTimeRangeFilters
 
-        if len(rawTimeRangeFilters) > 0:
-            return rawFilterString
-        else:
-            return None
+    def _history_compile_raw_filter(self, **options):
+        rawOptionsFilter = []
+        for key in options:
+            rawOptionsFilter.append(f"{key}%20eq%20'{options[key]}'")
+        return rawOptionsFilter
 
     def _get_history_by_resource_and_filter(self, resource_name, fromDate=None, toDate=None, **options):
-        rawTimeRangeFilter = self._history_timerange_filter_with_guard(fromDate=fromDate, toDate=toDate)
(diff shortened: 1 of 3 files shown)


balena/models/history.py
```diff
@@ -144,15 +149,15 @@ def get_all_by_application(self, app_id, fromDate=datetime.utcnow() + timedelta(
         Get all device history entries for an application.
 
         Args:
-            app_id (str): application id.
+            app_id: application id.
```
otaviojacobi: @fisehara Can we keep it as Union[str,int] type here? If eventually we manage to remove the types from the docs and into the python type hinting thing  this will make the transition smoother and will be compactible with the rest of the repo: https://github.com/balena-io/balena-sdk-python/blob/master/balena/models/tag.py#L209
```suggestion
            app_id (Union[str, int]): application id.
```
If I read this code correctly, you can pass either '2036095' or 2036095 and both should work as you do a cast to integer late, but please correct the type in case that's not it

balena/models/history.py
```diff
@@ -40,25 +41,26 @@ def _history_timerange_filter_with_guard(self, fromDate=None, toDate=None):
...
+        return rawTimeRangeFilters
 
-        if len(rawTimeRangeFilters) > 0:
-            return rawFilterString
-        else:
-            return None
+    def _history_compile_raw_filter(self, **options):
+        rawOptionsFilter = []
+        for key in options:
+            rawOptionsFilter.append(Template("$key%20eq%20'$value'").safe_substitute(key=key, value=options[key]))
```
otaviojacobi: @fisehara Just to know, why did you prefer Template + safe_substitute rather than
```suggestion
rawOptionsFilter.append(f"{key}%20eq%20'{options[key]}'")
```
? afaik they will both achieve the same thing

fisehara: @otaviojacobi I just copy/paste the existing procedure, and agree with your shortened version.

balena/models/history.py
fisehara: @otaviojacobi So I was confused as this is not completly consistent. So I guess the Union type is the more relaxed typing. I'll follow it that way.

otaviojacobi (approved):

PR #325, opened 2023-05-01
Update semver to 3.0.0

JSReds:
Update semver from 2.10.0 to 3.0.0

Change-type: patch

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,17 +8,17 @@ readme = "README.md"
 packages = [{include = "balena"}]
 
 [tool.poetry.dependencies]
-python = "^3.6.2"
+python = "^3.9.6"
 PyJWT = ">=2.0.0"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -8,17 +8,17 @@ readme = "README.md"
 packages = [{include = "balena"}]
 
 [tool.poetry.dependencies]
-python = "^3.6.2"
+python = "^3.9.6"
```
otaviojacobi, line 11: @JSReds i am not sure we can do this in a patch/if we can do this bump here. Basically we are deprecating support for python 3.6 3.7 and 3.8 in a patch bump. I am not sure we can do that as we have customers on python 3.7 (for sure) and I suspect some in 3.6

JSReds:
Closing in favour of https://github.com/balena-io/balena-sdk-python/pull/326

PR #326, opened 2023-05-05
SDK Version 13 with pine client and semver bump

otaviojacobi:
Change-type: major


balena/models/hup.py
```diff
@@ -15,39 +17,52 @@ def get_hup_action_type(self, device_type, current_version, target_version):
...
-            + self.__xstr(parsed_target_ver["build"])
-            + self.__xstr(parsed_current_ver["build"])
-        )
+        cur_variant = self.__get_variant(parsed_current_ver)
+        target_variant = self.__get_variant(parsed_target_ver)
 
-        if "dev" in releases:
-            raise exceptions.OsUpdateError("Updates cannot be performed on development balenaOS variants")
+        if target_variant is not None and (
+            (cur_variant == "dev") != (target_variant == "dev")
```
thgreasi: I would say let's split this tiny change out of this PR and maybe even ship it in the current major.

.gitignore
```diff
@@ -5,4 +5,5 @@ test.py
 build/*
 .vscode/
 dist/
-poetry.lock
\ No newline at end of file
+poetry.lock
+.coverage
```
pranasziaukas: Nit: I think it would be nice to have a newline at end of file, see [this](https://stackoverflow.com/questions/5813311/whats-the-significance-of-the-no-newline-at-end-of-file-log)

balena/auth.py
```diff
@@ -125,7 +125,6 @@ def who_am_i(self):
 
         Examples:
             >>> balena.auth.who_am_i()
-            u'g_trong_nghia_nguyen'
```
pranasziaukas: Just FYI, docstring pieces like that could be picked up and executed as tests automatically using [`doctest`](https://docs.python.org/3/library/doctest.html).

pranasziaukas: OFC this was never intended to work as a test so removing Trong's username seems like a good choice 👍

balena/dependent_resource.py
```diff
@@ -0,0 +1,88 @@
...
+            resource_key_field: str,
+            parent_resource_name: str,
+            get_resource_id: Callable[[Any], int]
+    ):
+        self.resource_name = resource_name
+        self.resource_key_field = resource_key_field
+        self.parent_resource_name = parent_resource_name
+        self.get_resource_id = get_resource_id
+
+    def _get_all(self, options: AnyObject = {}) -> List[T]:
```
pranasziaukas, line 25: Just be aware that mutable default arguments in python are considered a bad practice because it can lead to unintended (and confusing) side effects. For reference:
- https://www.codecademy.com/learn/learn-intermediate-python-3/modules/int-python-function-arguments/cheatsheet
- https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil

pranasziaukas: You are never mutating the argument so it should still be fine in practice. Up to you if you want to change/improve it.

balena/dependent_resource.py
```diff
@@ -0,0 +1,88 @@
...
+
+        return pine.get(
+            {
+                "resource": self.resource_name,
+                "options": merge(default_orderby, options),
+            }
+        )
+
+    def _get_all_by_parent(
+        self, parent_param: Any, options: AnyObject = {}
```
pranasziaukas: Is it possible to narrow down the `AnyObject` type hint here?
I think `Dict[str, Any]` should work and that would be more descriptive.

balena/dependent_resource.py
```diff
@@ -0,0 +1,88 @@
...
+                "body": {"value": value},
+            }
+        )
+
+    def _remove(self, parent_param: Any, tag_key: str) -> None:
+        parent_id = parent_param if is_id(parent_param) else self.get_resource_id(parent_param)
+
+        dollar_filter = {}
+        dollar_filter[self.parent_resource_name] = parent_id
+        dollar_filter[self.resource_key_field] = tag_key
```
pranasziaukas:
```suggestion
        dollar_filter = {
            self.parent_resource_name: parent_id,
            self.resource_key_field: key,
        }
```

^ IMO the above is cleaner

pranasziaukas: The same applies a couple of times in this file.

balena/exceptions.py
```diff
@@ -34,6 +34,23 @@ def __init__(self, option):
         self.message = Message.MISSING_OPTION.format(option=option)
 
 
+class SupervisorLocked(BalenaException):
+    """
+    Exception type for supervisor locked.
+
+    Args:
+        option (str): option name.
```
pranasziaukas: AFAICT the `option` argument is not used in the class.

balena/models/api_key.py
```diff
@@ -15,235 +20,169 @@ class ApiKey:
...
-        return self.base_request.request(
-            "api_key",
-            "GET",
-            params=filters,
-            endpoint=self.settings.get("pine_endpoint"),
-        )["d"]
 
-    def create_api_key(self, name, description=None):
+    def create(
+        self,
```
pranasziaukas, line 29: I noticed that more than half of the methods here are static, i.e. don't use the `self` reference.
However, I'm not sure if it's worth moving to `@staticmethod` at this point, especially if it has broader implications.

pranasziaukas: Might just be a note for the future.

balena/models/device.py
```diff
@@ -43,2025 +81,1423 @@ class DeviceStatus:
...
+                        "body": body,
+                    }
+                )
 
-            return None
         else:
-            raise exceptions.DeviceNotFound(uuid)
+            chunk_size = 200
+            chunked_devices = [
+                uuid_or_id_or_ids[i : i + chunk_size]  # noqa: E203 type: ignore
```
pranasziaukas, line 193: E203 is not part of the PEP8 standard, and ideally we should not be forcing it. But this is outside the scope of this PR OFC.

balena/models/device.py
```diff
@@ -43,2025 +81,1423 @@ class DeviceStatus:
...
+        return devices
 
-    def get_name(self, uuid):
+    def get_name(self, uuid_or_id: Union[str, int]) -> str:
         """
         Get device name by device uuid.
 
         Args:
-            uuid (str): device uuid.
+            uuid (str): device uuid (string) or id (int)
```
pranasziaukas:
```suggestion
            uuid_or_id (str): device uuid (string) or id (int)
```

balena/models/device.py
```diff
@@ -43,2025 +81,1423 @@ class DeviceStatus:
...
-        return self.base_request.request(
-            "blink",
-            "POST",
-            data=data,
-            endpoint=self.settings.get("api_endpoint"),
-            login=True,
-        )
-
-    def rename(self, uuid, new_name):
+    def rename(self, uuid_or_id_or_ids: Union[str, int], new_name: str) -> None:
```
pranasziaukas: Is this really meant to accept a list of device IDs? Otherwise
```suggestion
    def rename(self, uuid_or_id: Union[str, int], new_name: str) -> None:
```

balena/models/device.py
```diff
@@ -43,2025 +81,1423 @@ class DeviceStatus:
...
         """
 
-        device = self.get(uuid)
-        try:
-            self.__check_local_mode_supported(device)
-            return {"supported": True, "message": "Supported"}
-        except exceptions.LocalModeError as e:
-            return {"supported": False, "message": e.message}
+        if target_os_version is None or uuid is None:
+            raise exceptions.InvalidParameter("target_os_version or UUID", None)
```
pranasziaukas, line 1616: `exceptions.InvalidParameter` expects two string values according to its signature.

balena/models/device.py
```diff
@@ -43,2025 +81,1423 @@ class DeviceStatus:
...
-        else:
-            raw_data = raw_data[0]
+    def get_by_name(
+        self, name: str, options: AnyObject = {}
+    ) -> List[TypeDevice]:
+        """
+        Get devices by device name.
+
+        Args:
+            name (str): device name.
```
pranasziaukas, line 432:
```suggestion
            name (str): device name.
            options (AnyObject): extra pine options to use
```

balena/models/device.py
pranasziaukas: Might want to double-check other `uuid`/`uuid_or_id`/`uuid_or_id_or_ids` occurrences too :)

balena/models/device.py
```diff
@@ -43,2025 +81,1423 @@ class DeviceStatus:
...
+                    "is_of__cpu_architecture"
+                ][0]["slug"]
 
-        if device["current_services"]:
-            for service_name in device["current_services"]:
-                installs = device["current_services"][service_name]
-                install = max(installs, key=lambda x: x["id"])
-                if install and install["download_progress"] is not None and install["status"] == "Downloading":
-                    return DeviceStatus.UPDATING
+                err_msg = f"{device_type_slug} is not compactible with application {app_type_slug} device typ"
```
pranasziaukas, line 1184:
```suggestion
                err_msg = f"{device_type_slug} is not compatible with application {app_type_slug} device type"
```

balena/models/environment_variables.py
```diff
@@ -21,777 +19,435 @@ class EnvironmentVariable:
...
-        """  # noqa: E501
-
-        raw_query = "$filter=device/any(d:d/uuid%20eq%20'{uuid}')".format(uuid=uuid)
-        return self.base_request.request(
-            "device_environment_variable",
-            "GET",
-            raw_query=raw_query,
-            endpoint=self.settings.get("pine_endpoint"),
-        )["d"]
+            >>> balena.models.environment_variables.device.get_all_by_device(8deb12a)
```
pranasziaukas:
```suggestion
            >>> balena.models.environment_variables.device.get_all_by_device("8deb12a")
```

balena/models/environment_variables.py
```diff
@@ -21,777 +19,435 @@ class EnvironmentVariable:
...
-        )
+        if (
+            service_installs is None
+            or (
+                isinstance(service_installs, list)
+                and len(service_installs) == 0
+            )
+            or service_installs[0] is None
+        ):
+            raise exceptions.ServiceNotFound(service_id)
```
pranasziaukas: `exceptions.ServiceNotFound` expects a string according to its signature. `service_id` is an integer.

balena/models/environment_variables.py
```diff
@@ -21,777 +19,435 @@ class EnvironmentVariable:
...
         """
-        Get all device environment variables for an application.
+        Remove a device environment variable.
 
         Args:
-            app_id (str): application id.
-
-        Returns:
-            list: list of device environment variables.
+            uuid_or_id (Union[str, int]): device uuid (string) or id (int)
```
pranasziaukas:
```suggestion
            uuid_or_id (Union[str, int]): device uuid (string) or id (int)
            key (str): a variable name to be removed
```

balena/models/environment_variables.py
```diff
@@ -21,777 +19,435 @@ class EnvironmentVariable:
...
-
-    def remove(self, var_id):
+    def remove(self, slug_or_uuid_or_id: Union[str, int], key: str) -> None:
         """
-        Remove service environment variable.
+        Remove an application environment variable.
 
         Args:
-            var_id (str): service environment variable id.
+            slug_or_uuid_or_id (Union[str, int]): application slug (string), uuid (string) or id (number)
```
pranasziaukas:
```suggestion
            slug_or_uuid_or_id (Union[str, int]): application slug (string), uuid (string) or id (number)
            key (str): a variable name to be removed
```

balena/models/os.py
```diff
@@ -0,0 +1,80 @@
...
+        pass
+
+    def get_config(
+        self, slug_or_uuid_or_id: Union[str, int], options: ImgConfigOptions
+    ):
+        """
+        Download application config.json.
+
+        Args:
+            app_id (str): application id.
```
pranasziaukas: Doesn't exist

balena/models/release.py
```diff
@@ -71,13 +84,21 @@ def __get_by_raw_query(self, raw_query):
...
+        options: AnyObject = {},
+    ) -> AnyObject:
         """
         Get a specific release.
 
         Args:
-            commit_or_id: release commit (str) or id (int).
-
+            commit_or_id_or_raw_version(Union[str, int, ReleaseRawVersionApplicationPair]): release commit (string) or
+            id (number) or an object with the unique `application` (number or string) & `rawVersion` (string)
```
pranasziaukas: Doesn't exist

balena/models/tag.py
```diff
@@ -21,515 +20,437 @@ def __init__(self):
...
-                },
-            ]
-
         """
 
-        return super(DeviceTag, self).get_all()
+        return super(DeviceTag, self)._get_all(options)
 
-    def set(self, parent_id, tag_key, value):
+    def get(self, uuid_or_id: Union[str, int], tag_key: str) -> Optional[str]:
```
pranasziaukas: Please remove `value` from the docstring as well

balena/pine.py
```diff
@@ -0,0 +1,49 @@
...
+
+        headers = {"Content-Type": "application/json"}
+        if token is not None:
+            headers["Authorization"] = f"Bearer {token}"
+
+        data = None
+        if body is not None:
+            data = json.dumps(body)
+
+        req = requests.request(method, url=url, data=data, headers=headers)
```
pranasziaukas, line 38: Fun fact: `requests.request` returns a `request.Response` object, so I would be tempted to name the resulting variable that way too. However, your naming choice is also adopted in the official documentation so no complaints.

balena/pine.py
```diff
@@ -0,0 +1,49 @@
...
+from pine_client import PinejsClientCore
+from pine_client.client import Params
+
+from .balena_auth import get_token
+from .exceptions import RequestError
+from .settings import settings
+
+
+class PineClient(PinejsClientCore):
+    def __init__(self, params: Params = {}):
```
pranasziaukas: Since this is a new class, I would not recommend to use mutable default arguments here (see my comment above).

```suggestion
    def __init__(self, params: Params = None):
        if params is None:
            params = {}
```

These are two extra lines but it's a worthwhile tradeoff from my personal experience.

pranasziaukas: We can discuss this pattern more if you'd like.

balena/pine.py
```diff
@@ -0,0 +1,49 @@
...
+        data = None
+        if body is not None:
+            data = json.dumps(body)
+
+        req = requests.request(method, url=url, data=data, headers=headers)
+
+        if req.ok:
+            try:
+                return req.json()
+            except Exception:
```
pranasziaukas, line 43: Ideally, we'd try to handle specific exceptions instead of catching a generic `Exception` here. You think it's possible?

balena/settings.py
```diff
@@ -199,8 +201,6 @@ def set(self, key, value):
 
     def remove(self, key):
         """
-        Remove a setting.
```
pranasziaukas, line 202: Can't say if it's the best approach to remove the docstring description here altogether. I suppose your reasoning was that it is self-explanatory?

pyproject.toml
```diff
@@ -8,17 +8,19 @@ readme = "README.md"
 packages = [{include = "balena"}]
 
 [tool.poetry.dependencies]
-python = "^3.6.2"
+python = "^3.8.1"
```
pranasziaukas, line 11: Great stuff!

balena/utils.py
```diff
@@ -46,3 +56,182 @@ def compare(a, b):
...
+def with_supervisor_locked_error(fn: Callable[[], T]) -> T:
+    try:
+        return fn()
+    except RequestError as e:
+        if e.status_code == SUPERVISOR_LOCKED_STATUS_CODE:
+            raise SupervisorLocked()
+        raise e
+
+
+def normalize_balena_semver(os_version):
```
pranasziaukas: AFAICT this has been copy-pasted from the `balena/models/device_os.py` file. Should the original function be deleted?

balena/utils.py
```diff
@@ -46,3 +56,182 @@ def compare(a, b):
...
+    raw_data.pop("is_provided_by__release", None)
+    raw_data.pop("image", None)
+    install.update(raw_data)
+
+    return install
+
+
+def generate_current_service_details(raw_device: Any) -> Any:
+    # TODO: Please compare me to node-sdk version
+    groupedServices = defaultdict(list)
```
pranasziaukas: Snake case is preferred it python. What do you think about
```suggestion
    grouped_services = defaultdict(list)
```

balena/utils.py
```diff
@@ -1,22 +1,32 @@
...
+def is_id(value: Any) -> bool:
     """
     Return True, if the input value is a valid ID. False otherwise.
 
     """
 
     if isinstance(value, numbers.Number):
         try:
-            int(value)
+            int(value)  # type: ignore
```
pranasziaukas, line 21: What is this about?

balena/utils.py
pranasziaukas: Also, it's the only function w/o type hints in the file. Would be awesome to add it for the sake of consistency.

pranasziaukas (commented):
A pretty colossal work here Otávio.

I have only done sporadic API work so far from an expert here. My review is more about pythonic observations, some best practices, etc. Hope you find it useful anyway.

Cheers!

balena/dependent_resource.py
otaviojacobi: @pranasziaukas after too many hours reading between TS and Python code I somehow totally forgot I could do that in python as this is a dict and not a TS obejct. thanks!

balena/models/api_key.py
otaviojacobi: Well, that was my first thought when reading this repo, but if we think it thought actually `everything` here is a @staticmethod, even the ones that depend on an instance of self.application for example, as each model could be a singleton, everything could be static (and somehow is, nothing preserves state really, we just use properties belonging to self to store some constants etc, which can also be improved)

My take was not to use @staticmethod on everything to avoid making this PR even huger than it is, but it is for sure something that can be discussed

balena/dependent_resource.py
otaviojacobi: @pranasziaukas AnyObject is exactly:  `AnyObject = Dict[str, Any]` and it is owned by us, it is mostly a translation from AnyObject from node-sdk. But I do plan on minimizing its usage with better type hints in minors later

balena/models/release.py
otaviojacobi: I guess this is a formatting problem, this is just the continuation of the previous line

balena/pine.py
otaviojacobi: Yes, I agree we can do a a better job handling exceptions here in general. But I think it should go into the todo list for minors following

balena/settings.py
otaviojacobi: This was a miss deletion :)

balena/utils.py
otaviojacobi: I think that somehow flake8 is not screaming at me for some reason, I will look into it, but for sure we should be just using just snake_case

balena/utils.py
otaviojacobi: Yep, I am still refactoring device_os local and the types and deduplication is coming on the next commit :)

pyproject.toml
otaviojacobi: Just out of fun you might like: https://github.com/balena-io/balena-sdk-python/pull/320/files :P

.gitignore
otaviojacobi: @lmbarros will agree with you :)

balena/auth.py
otaviojacobi: Yeah, overall I would like to remove the examples from the documentations as everytime that api does some change they get obsolate and not necessarily the SDK tests will run when API changes so it makes it very hard to mantain. My bet here is that it will be a better developer experience and easier to mantain (as we can basically just map changes on node-sdk types) if we replace it by type hints.

otaviojacobi: What we might want to look into in the future (for both node and python sdks) is generating the examples from a job that runs when balena-api/open-balena-api is merged. For example, the outputs from e2e tests or something. This would be much more valuable imho

balena/models/api_key.py
pranasziaukas: > My take was not to use `@staticmethod` on everything to avoid making this PR even huger than it is, but it is for sure something that can be discussed

Indeed, that was my line of thinking too!

docs/doc2md.py
```diff
@@ -297,5 +396,29 @@ def add_path(*pathes):
...
 
 
+def typed_dict_to_dict(typed_dict_cls):
+    return {k: simplify_type_hint(v) for k, v in get_type_hints(typed_dict_cls).items()}
+
+
+def pretty_print_python_dict(d):
+    items = [f'    "{k}": {v}' for k, v in d.items()]
+    p_dict = "{\n" + ",\n".join(items) + "\n}"
+    print(p_dict)
```
fisehara: I'd give the variables k and v proper names. From my understanding k is the propertyName and v is the propertyType

docs/doc2md.py
```diff
@@ -98,21 +147,71 @@ def find_sections(lines):
...
+
         ref = ref.replace(" ", "-")
         ref = ref.replace("?", "")
         refs.append(INDENT * (ind) + "- [%s](#%s)" % (sec, ref))
+
+        for (son_name, hint, son_ref) in sons:
+            hint_ref = None
+            for h in hints:
+                if h in hint:
+                    hint_ref = h.lower()
```
fisehara: So I understand that `hints` is an array containing strings (str) and `hint` is a single string (str) coming from `simplify_type_hint`
Wouldn't then this be enoug:
```suggestion
            hint_ref = None
            if hint in hints:
                hint_ref = hint.lower()
```

docs/doc2md.py
```diff
@@ -5,6 +5,55 @@
...
+
+
+def simplify_type_hint(type_hint: Any) -> str:
+
+    # print(type_hint)
+
+    if isinstance(type_hint, str):
+        return f'"{type_hint}"'
+    elif type_hint is None or type_hint.__name__ == "NoneType":
+        # print("A")
```
fisehara: debug information?
See following lines

docs/doc2md.py
```diff
@@ -5,6 +5,55 @@
...
+    elif type_hint is Any:
+        # print("C")
+        return 'Any'
+    elif hasattr(type_hint, '_name') and type_hint._name:
+        # print("D")
+
+        if not hasattr(type_hint, "__args__"):
+            return type_hint._name
+
+        inner_types = [simplify_type_hint(t) for t in type_hint.__args__]
```
fisehara: I don't know if we have a solid style guide in puython projects, but I'd like to suggest to use readable variable names, instead of single character ones.

docs/doc2md.py
```diff
@@ -98,21 +147,71 @@ def find_sections(lines):
...
     Generate table of contents for array of section names.
     """
+
     if not sections:
         return []
     refs = []
-    for sec, ind in sections:
-        ref = sec.lower()
+    for sec, ind, ref in sections:
+        sons = []
```
fisehara: I understand that this refers to the `children` of a node. I'd suggest calling it children as this is familiar for graph structures. I'm hickuping reading sons :)

docs/doc2md.py
```diff
@@ -98,21 +147,71 @@ def find_sections(lines):
...
+            if hint_ref:
+                refs.append(INDENT * (ind+1) + f"- [{son_name}](#{son_ref}) ⇒ [<code>{hint}</code>](#{hint_ref})")
+            else:
+                refs.append(INDENT * (ind+1) + f"- [{son_name}](#{son_ref}) ⇒ <code>{hint}</code>")
+
     return "\n".join(refs)
 
 
+def get_funcs(baseclass):
+    sons = []
```
fisehara:
```suggestion
    children = []
```

docs/doc2md.py
```diff
@@ -98,21 +147,71 @@ def find_sections(lines):
...
+    for func_name, _ in inspect.getmembers(baseclass, predicate=inspect.isfunction):
+        if func_name != "__init__" and not func_name.startswith("_"):
+            func = getattr(baseclass, func_name)
+            if "self" in inspect.getfullargspec(func)[0]:
+                print_name, hint = make_function_name(func, func_name)
+                ref_name = f"{baseclass.__name__.lower()}.{func_name}"
+
+                sons.append((print_name, hint, ref_name))
+    return sons
```
fisehara:
```suggestion
                children.append((print_name, hint, ref_name))
    return children
```

docs/doc2md.py
```diff
@@ -98,21 +147,71 @@ def find_sections(lines):
...
+def make_function_name(func, func_name):
+    args_list = inspect.getfullargspec(func)[0]
+    if "self" in args_list:
+        args_list.remove("self")
+
+    f_args = ", ".join(args_list)
+
+    fqh = get_type_hints(func).get("return")
+    if fqh:
+        hint = simplify_type_hint(fqh)
```
fisehara: What dos fqh stand for? fullyQualifiedHint? Can you please use better readabile var names.

docs_generator.py
```diff
@@ -61,171 +59,218 @@ def print_newline():
...
             func = getattr(baseclass, func_name)
-            print(doc2md.doc2md(func.__doc__, make_function_name(func, func_name), type=1))
+            print(f'\n<a name="{baseclass.__name__.lower()}.{func_name}"></a>')
+
+            print_name, hint = doc2md.make_function_name(func, func_name)
+
+            hint_ref = None
+            for h in hints:
+                if h in hint:
+                    hint_ref = h.lower()
```
fisehara:
```suggestion
            hint_ref = None
            if hint in hints:
                hint_ref = hint.lower()
```

tests/helper.py
```diff
@@ -86,335 +77,217 @@ def wipe_application(self):
...
-        params = {"filter": "1", "eq": 1}
-
-        self.balena.models.application.base_request.request(
-            "application",
-            "DELETE",
-            params=params,
-            endpoint=self.balena.settings.get("pine_endpoint"),
-            login=True,
-        )
+        self.balena.pine.delete({"resource": "application", "options": {"filter": "1", "eq": 1}})
```
fisehara: Are `"filter"` and `"eq"` allowed without `$` prefix? For OData 4.0 it's optional to give the `$` but for pinejs and it's ODate 3.0 compatible implementation I recall the `$` to be mandatory.

tests/helper.py
```diff
@@ -86,335 +77,217 @@ def wipe_application(self):
...
             self.wipe_application()
-            self.balena.models.key.base_request.request(
-                "user__has__public_key",
-                "DELETE",
-                params=params,
-                endpoint=self.balena.settings.get("pine_endpoint"),
-                login=True,
-            )
+
+            self.balena.pine.delete({"resource": "user__has__public_key", "options": {"filter": "1", "eq": 1}})
```
fisehara: `$` skipped for OData parameters

tests/functional/test_logs.py
```diff
@@ -0,0 +1,179 @@
...
+                "message": 'First message',
+                "timestamp": int(time.time() * 1000)
+            },
+            {
+                "message": 'Second message',
+                "timestamp": int(time.time() * 1000)
+            },
+        ])
+
+        time.sleep(2)
```
fisehara, line 62: Would it otherwise cause a raise condition? If the POST logs returns the system should have persistet them, guaranteed

tests/functional/test_logs.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def test_05_subscribe_should_stream_new_logs(self):
+        results = []
+
+        def cb(data):
+            results.append(data["message"])
+
+        self.balena.logs.subscribe(self.uuid, cb)
+
+        time.sleep(0.1)
```
fisehara: For what is thie needed?

tests/functional/test_logs.py
```diff
@@ -0,0 +1,179 @@
...
+                "message": 'First message',
+                "timestamp": int(time.time() * 1000)
+            },
+            {
+                "message": 'Second message',
+                "timestamp": int(time.time() * 1000)
+            },
+        ])
+
+        time.sleep(2)
```
fisehara: Same here, for what is this needed? If you subscribe, the subscription should wait until receive, otherwise the testcase times out and fails, no?

tests/functional/test_logs.py
```diff
@@ -0,0 +1,179 @@
...
+        self.assertEqual(results, [])
```
fisehara: Please revisit all time.sleep(xyz) and make sure that an awaited syncronuouse data flow guarantees that the data is there before requesting the logs.

tests/functional/test_auth.py
```diff
@@ -0,0 +1,139 @@
...
+    def test_05_should_reject_to_get_token_when_not_logged_in(self):
+        self.assertIsNone(self.balena.auth.get_token())
+
+    def test_06_should_be_able_to_login_with_session_token(self):
+        token = self.balena.auth.authenticate(**TestAuth.creds)
+        self.balena.auth.login_with_token(token)
+        self.assertEqual(self.balena.auth.get_token(), token)
+
+    def test_07_should_be_able_to_login_with_api_key(self):
+        key = self.balena.models.api_key.create("apiKey")
```
fisehara, line 46: Should login with credentials and not rely on the previous login.
```suggestion
        self.balena.auth.authenticate(**TestAuth.creds)
        key = self.balena.models.api_key.create("apiKey")
```

tests/functional/test_auth.py
```diff
@@ -0,0 +1,139 @@
...
+
+        user_id = self.balena.auth.get_user_id()
+        self.assertIsInstance(user_id, int)
+        self.assertGreater(user_id, 0)
+
+        user_actor_id = self.balena.auth.get_user_actor_id()
+        self.assertIsInstance(user_actor_id, int)
+        self.assertGreater(user_actor_id, 0)
+
+    def test_14_should_be_able_to_login(self):
```
fisehara: could also be checked in test_08
```suggestion
    def test_14_should_not_return_logged_in_when_logged_out(self):
```

tests/helper.py
```diff
@@ -86,335 +77,217 @@ def wipe_application(self):
...
+        old_web_image = self.balena.pine.post(
+            {
+                "resource": "image",
+                "body": {
+                    "is_a_build_of__service": web_service["id"],
+                    "project_type": "dockerfile",
+                    "content_hash": "abc",
+                    "build_log": "old web log",
+                    "start_timestamp": 1234,
+                    "push_timestamp": 1234,
```
fisehara: Would it potentially make more sense to generate an actual timestamp for these fixtures?

tests/functional/models/test_release.py
```diff
@@ -0,0 +1,394 @@
...
+        cls.helper = TestHelper()
+        cls.balena = cls.helper.balena
+        cls.helper.wipe_application()
+        cls.empty_app = cls.balena.models.application.create(
+            "ServiceTestApp", "raspberry-pi2", cls.helper.default_organization["id"]
+        )
+        cls.mc_app = cls.helper.create_multicontainer_app()
+        cls.TEST_SOURCE_URL = "https://github.com/balena-io-examples/balena-node-hello-world/archive/v1.0.0.tar.gz"
+        cls.TEST_SOURCE_CONTAINER_COUNT = 1
+        cls.UNIQUE_PROPERTY_NAMES = ["id", "commit", "__belongs_to__hash__"]
```
fisehara: Should the `__belongs_to__hash__` be modelled differently? It's not an balena API property name.

tests/functional/models/test_release.py
```diff
@@ -0,0 +1,394 @@
...
+                        "start_timestamp": datetime.datetime.now().isoformat(),
+                    },
+                }
+            )
+            TestRelease.test_release_by_field[field] = self.balena.models.release.get(
+                TestRelease.test_release_by_field[field]["id"],
+                {"$select": ["id", "commit", "raw_version", "belongs_to__application"]},
+            )
+
+        for field in self.UNIQUE_PROPERTY_NAMES:
```
fisehara: Can't this run in the same loop as above?

tests/functional/models/test_release.py
```diff
@@ -0,0 +1,394 @@
...
+        self.assertEqual(release["commit"], "new-release-commit")
+        self.assertEqual(release["id"], match["id"])
+        self.assertEqual(release["belongs_to__application"]["__id"], match["app_id"])
+
+    def __expect_release_to_match_on_get_with_image_details(self, release, build_logs=[None, None]):
+        self.assertEqual(release["commit"], "new-release-commit")
+        self.assertEqual(release["status"], "success")
+        self.assertEqual(release["source"], "cloud")
+        self.assertEqual(release["images"][0]["service_name"], "db")
+        self.assertEqual(release["images"][1]["service_name"], "web")
```
fisehara, line 352: This may fail if the order isn't correct by any chance?

tests/functional/models/test_release.py
```diff
@@ -0,0 +1,394 @@
...
+        self.assertEqual(release["user"]["username"], self.helper.credentials["user_id"])
+        self.assertEqual(release["images"][0].get("build_log"), build_logs[0])
+        self.assertEqual(release["images"][1].get("build_log"), build_logs[1])
+
+    def __get_param(self, field: str, draft_release: Any):
+        if field == "__belongs_to__hash__":
```
fisehara: Can you add some details what this `__belongs_to__hash__` is?

tests/functional/models/test_organization.py
```diff
@@ -107,26 +99,73 @@ def test_invite_get_all(self):
...
+        if value is not None:
+            self.assertEqual(1, len(membership_tag_model.get_all_by_organization(org_id)))
+            self.assertEqual(
+                1,
+                len(membership_tag_model.get_all_by_organization_membership(membership_id)),
+            )
+
+        self.assertEqual(membership_tag_model.get(membership_id, key), value)
+
+        if value is not None:
```
fisehara: Can both if statement blocks be combined?

tests/functional/models/test_environment_variables.py
```diff
@@ -0,0 +1,235 @@
...
+                f"VIM{resource}_edit",
+            )
+
+    def test_05_can_remove_application_env_vars(self):
+        for resource in self.__app_fetch_resources:
+            self.app_env_var.remove(self.app[resource], f"EDITOR_BY_{resource}")
+            self.assertIsNone(
+                self.app_env_var.get(self.app[resource], f"EDITOR_BY_{resource}"),
+            )
+
```
fisehara, line 76: For completeness you may also want to add tests for `TestApplicationServiceEnvironmentVariables`

fisehara (commented):
I'm not finished, but wanted to give this part review already

docs/doc2md.py
otaviojacobi: > I'm not finished, but wanted to give this part review already

@fisehara  A good chunk of the comments here are on doc2md/docs_generator which I admit I treated as second class. I will do a double check on the whole thing and address your comments

docs/doc2md.py
otaviojacobi: No @fisehara, I will change the names to make this more explicitely.
For each h in the hints list, h will be a hint (for example MyType, of all types declared in models.py) and "hint" is the actual string hint of that child, for example List[MyType], which we would still like to point to MyType, so this is a string comparison to check if it is a subset.

docs_generator.py
otaviojacobi: Same as above, this is not the same. I fixed with better naming and a small explaining comment

tests/helper.py
otaviojacobi: very nice catch. I wonder how is this even working, passing... I guess the filter is just not applying :thinking:

tests/functional/test_auth.py
otaviojacobi: @fisehara I can fix this one but I would like to let you know that these tests are not indepedent. That's why they have the test_{number} prefix. The main reason for that is we could do tests that are fully independent but this slows down our pipeline a lot. Thodoris did the first version where the result of one test can be used on the next one and I agree it is the best we can get here, otherwise deployments get too slow (2+ hours)

tests/functional/test_auth.py
otaviojacobi: yep, copy paste error. thanks

tests/functional/models/test_release.py
otaviojacobi: @fisehara I agree we might want to modify the constant name but the things is, a dict is not a hashable parameter, this basically follows the node-sdk pattern (which does similar)

tests/helper.py
otaviojacobi: I don't see a real gain from it tbh, but it is such a simple change that I did it anyway

tests/functional/models/test_release.py
otaviojacobi: I will add some explanation

tests/functional/models/test_environment_variables.py
otaviojacobi: hey @fisehara thanks for the heads up, this one was already in my list (this repo has overall less test coverage than the node-sdk one, I am still in the process of increasing it)

tests/helper.py
fisehara: @otaviojacobi yeah the filter is not applying at all.

tests/functional/models/test_environment_variables.py
otaviojacobi: done

tests/functional/models/test_release.py
otaviojacobi: AFAIK, this also tests the order which we are expecting the returned images to be in.

tests/functional/test_logs.py
otaviojacobi: @fisehara I am not sure we can have such a guarantee, because it depends on the time the API will be able to handle and forward the logs. The node-sdk does similar: https://github.com/balena-io/balena-sdk/blob/master/tests/integration/logs.spec.ts#L61

.gitignore
```diff
@@ -5,4 +5,5 @@ test.py
 build/*
 .vscode/
 dist/
-poetry.lock
\ No newline at end of file
+poetry.lock
```
pranasziaukas, line 8: While not required, the default practice is to commit the lock file so that users install the exact same version of dependencies. See https://python-poetry.org/docs/basic-usage#installing-without-poetrylock

The same applies to `package-lock.json` in the NPM world, and AFAICT we tend to commit it in our repos.

balena/balena_auth.py
```diff
@@ -0,0 +1,91 @@
...
+    else:
+        api_key = os.environ.get("BALENA_API_KEY") or os.environ.get("RESIN_API_KEY")
+
+    return api_key
+
+
+def request(
+    method: str,
+    path: str,
+    body: Optional[Any] = None,
```
pranasziaukas, line 57: If it's a serialized JSON then maybe
```suggestion
    body: Optional[str] = None,
```

balena/balena_auth.py
```diff
@@ -0,0 +1,91 @@
...
+    endpoint: Optional[str] = None,
+    token: Optional[str] = None,
+    qs: Optional[Any] = {},
```
pranasziaukas, line 60: Similarly, it would be awesome to have something more informative than `Any`, if possible.

pranasziaukas: Also, a mutable default argument concern we discussed previously (https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments, etc.)

balena/balena_auth.py
```diff
@@ -0,0 +1,91 @@
...
+
+        try:
+            return req.json()
+        except Exception:
+            return req.content.decode()
+
+    except Exception as e:
+        if not send_token:
+            raise e
+        raise exceptions.NotLoggedIn()
```
pranasziaukas, line 95: Would we want to output anything from the original generic exception `e`? Or can we safely conclude that `exceptions.NotLoggedIn` is enough?

balena/dependent_resource.py
```diff
@@ -0,0 +1,76 @@
...
+        default_orderby = {"$orderby": {self.resource_key_field: "asc"}}
+
+        return pine.get(
+            {
+                "resource": self.resource_name,
+                "options": merge(default_orderby, options),
+            }
+        )
+
+    def _get_all_by_parent(self, parent_param: Any, options: AnyObject = {}) -> List[T]:
```
pranasziaukas, line 35: Would it be possible to improve the generic `Any` type by using `Dict`, `Union`, etc. types?

For example, `Dict[str, Any]` or `Union[SomeClass, Dict[str, Any]]` are both better (more informative) in my eyes.

balena/dependent_resource.py
```diff
@@ -0,0 +1,76 @@
...
+        dollar_filter = {self.parent_resource_name: parent_id, self.resource_key_field: key}
+
+        result = pine.get(
+            {
+                "resource": self.resource_name,
+                "options": {"$select": "value", "$filter": dollar_filter},
+            }
+        )
+
+        if len(result) == 1:
```
pranasziaukas: Do we really want to ignore in case of more than one result?

Maybe
```suggestion
        if len(result) >= 1:
```

balena/dependent_resource.py
```diff
@@ -0,0 +1,76 @@
...
+        parent_id = parent_param if is_id(parent_param) else self.get_resource_id(parent_param)
+
+        get_options = {
+            "$filter": {self.parent_resource_name: parent_id},
+            "$orderby": f"{self.resource_key_field} asc",
+        }
+
+        return self._get_all(merge(get_options, options))
+
+    def _get(self, parent_param: Any, key: str) -> Optional[str]:
```
pranasziaukas, line 45: This uses `key` whereas the next two methods use `tag_key` to point to the same `resource_key_field` attribute.

Should they be the same for the sake of consistency?

balena/exceptions.py
```diff
@@ -583,3 +598,17 @@ class BalenaInvalidDeviceType(BalenaException):
     def __init__(self, device_type):
         super(BalenaInvalidDeviceType, self).__init__()
         self.message = Message.BALENA_INVALID_DEVICE_TYPE.format(device_type=device_type)
+
+
+class TooManyRequests(BalenaException):
```
pranasziaukas, line 603: Great additions in this file 👍

balena/logs.py
```diff
@@ -46,12 +45,13 @@ def dataReceived(self, data):
                 if line:
                     obj = json.loads(line)
             except Exception as e:
-                self.transport.stopProducing()
-                self.transport.loseConnection()
+                self.transport.stopProducing()  # type: ignore
+                self.transport.loseConnection()  # type: ignore
```
pranasziaukas, line 50: What is this about?

balena/models/api_key.py
```diff
@@ -13,237 +18,156 @@ class ApiKey:
...
+        if description is not None and isinstance(description, str):
+            api_key_body["description"] = description
 
-        return self.base_request.request(
-            "/api-key/user/full",
-            "POST",
-            data=data,
-            endpoint=self.settings.get("api_endpoint"),
-        )
+        if expiry_date is not None and isinstance(expiry_date, str):
```
pranasziaukas, line 55: Would it benefit from an additional check to make sure the date follows a valid format?

balena/models/api_key.py
```diff
@@ -13,237 +18,156 @@ class ApiKey:
...
         Returns:
-            str: user API key.
+            str: API key
 
         Examples:
-            >>> balena.models.api_key.create_api_key('myApiKey')
-            '3YHD9DVPLe6LbjEgQb7FEFXYdtPEMkV9'
-
+            >>> balena.models.api_key.create_api_key("myApiKey")
+            >>> balena.models.api_key.create_api_key("myApiKey", "my api key description")
```
pranasziaukas, line 47: Should it have an example with an `expiry_date` as well since it was added to the function?

balena/models/application.py
```diff
@@ -1,533 +1,427 @@
...
-            "service_id": service["id"],
-        }
+        Examples:
+            >>> balena.models.application.get_dashboard_url(1476418)
+        """
+        try:
+            if isinf(int(app_id)):
+                raise exceptions.InvalidParameter("app_id", app_id)
+        except ValueError:
+            raise exceptions.InvalidParameter("app_id", app_id)
```
pranasziaukas, line 143: This particular method features a nice ID sanity check. Should this check be refactored into a small helper method and used whenever an ID is passed?

balena/models/billing.py
```diff
@@ -0,0 +1,268 @@
...
+
+        Args:
+            organization (Union[str, int]): handle (string) or id (number) of the target organization.
+            plan_change_options (PlanChangeOptions): a dict containing billing account info
+
+        Examples:
+            >>> balena.models.billing.change_plan('myorghandle', { billingCode: 'prototype-v2', cycle: 'annual' })
+        """
+        org_id = self.__get_org_id(organization)
+        body = {**plan_change_options, "annual": plan_change_options["cycle"] == "annual"}
```
pranasziaukas, line 241: Should we validate the value of `cycle` to prevent mistypes, etc.?

balena/models/billing.py
```diff
@@ -0,0 +1,268 @@
...
+        Args:
+            organization (Union[str, int]): handle (string) or id (number) of the target organization.
+
+        Examples:
+            >>> balena.models.billing.get_invoices('myorghandle')
+        """
+        org_id = self.__get_org_id(organization)
+        return request(method="GET", path=f"/billing/v1/account/{org_id}/invoices")
+
+    def download_invoice(self, organization: Union[str, int], invoice_number: Union[str, int]):
```
pranasziaukas, line 257: This is the only method missing a return type annotation.

balena/models/credit_bundle.py
```diff
@@ -0,0 +1,66 @@
...
+                "resource": "credit_bundle",
+                "options": merge(
+                    options, {"$filter": {"belongs_to__organization": org_id}, "$orderby": {"created_at": "desc"}}
+                ),
+            }
+        )
+
+    def create(self, organization: Union[str, int], feature_id: int, credits_to_purchase: float) -> CreditBundleType:
+        """
+        Get all of the credit bundles purchased by the given org
```
pranasziaukas: Pretty sure this docstring is incorrect :) A copy-paste error?

balena/models/credit_bundle.py
```diff
@@ -0,0 +1,66 @@
...
+
+class CreditBundle:
+    """
+    This class implements credit bundle model for balena python SDK.
+    """
+
+    def __get_org_id(self, organization: Union[str, int]) -> int:
+        return org_model.get(organization, {"$select": "id"})["id"]
+
+    def get_all_by_org(self, organization: Union[str, int], options: AnyObject = {}) -> List[CreditBundleType]:
```
pranasziaukas, line 22: Mutable default argument :P

balena/models/device.py
```diff
@@ -43,2025 +98,2247 @@ class DeviceStatus:
...
-                params=params,
-                endpoint=self.settings.get("pine_endpoint"),
-            )["d"]
-            if len(devices) > 1:
-                raise exceptions.AmbiguousDevice(uuid)
-            return devices[0]
-        except IndexError:
-            raise exceptions.DeviceNotFound(uuid)
+        get_hup_action_type(
+            device_info["is_of__device_type"][0]["slug"],
```
pranasziaukas, line 243: We are accessing a dictionary 3-levels-deep. What is the risk of structure changing and causing an exception?

[`dict.get`](https://docs.python.org/3/library/stdtypes.html#dict.get) might help to mitigate the risk.

balena/models/device.py
```diff
@@ -43,2025 +98,2247 @@ class DeviceStatus:
...
+            uuid_or_id,
+            {
+                "$select": "id",
+                "$expand": {"belongs_to__application": {"$select": "app_name"}},
+            },
+        )
+        return device["belongs_to__application"][0]["app_name"]
 
-    def has(self, uuid):
+    def has(self, uuid_or_id: Union[str, int]) -> bool:
```
pranasziaukas, line 482: Not sure if `has` is a descriptive function name. Its docstring talks about device existence. Maybe the function should be named `exists` instead?

balena/models/history.py
```diff
@@ -1,209 +1,130 @@
...
+        to_date: Optional[datetime] = None,
+        options: AnyObject = {},
+    ) -> List[DeviceHistoryType]:
         """
         Get all device history entries for a device.
 
         Args:
-            uuid_or_id (str): device uuid or id.
+            uuid_or_id (str): device uuid (32 / 62 digits string) or id (number) __note__: No short IDs supported
+            from_date (datetime): history entries older or equal to this date. Defaults to 7 days ago
```
pranasziaukas: The phrasing "entries older or equal to this date" sounds unnatural to me.

How about something like "entries newer than or equal to this timestamp" or similar?

balena/models/history.py
```diff
@@ -1,209 +1,130 @@
...
+            from_date (datetime): history entries younger or equal to this date.
```
pranasziaukas:
```suggestion
            to_date (datetime): history entries younger or equal to this date.
```

balena/models/history.py
```diff
@@ -1,209 +1,130 @@
...
+        options: AnyObject = {},
+    ) -> List[DeviceHistoryType]:
         """
         Get all device history entries for an application.
 
         Args:
-            app_id (Union[str, int]): application id.
+            slug_or_uuid_or_id (Union[str, int]): application slug (string), uuid (string) or id (number)
+            from_date (datetime): history entries older or equal to this date. Defaults to 7 days ago
+            from_date (datetime): history entries younger or equal to this date.
```
pranasziaukas: Same comment as with `get_all_by_device`

balena/models/key.py
```diff
@@ -13,89 +14,63 @@ class Key:
...
+        Args:
+            options (AnyObject): extra pine options to use
 
+        Returns:
+            List[SSHKeyType]: list of ssh keys.
         """
-        return self.base_request.request("user__has__public_key", "GET", endpoint=self.settings.get("pine_endpoint"))[
-            "d"
-        ]
+        return pine.get({"resource": "user__has__public_key", "options": options})
```
pranasziaukas: `"user__has__public_key"` string is used 4 times in this relatively small class. In case this ever changes, you would have to find and replace it in several places. An arguably better approach would be to define it as a constant at the beginning of your class.

This comment is also applicable for quite a few other classes :)

balena/models/organization.py
```diff
@@ -410,276 +237,165 @@ class OrganizationMembership:
...
-        if len(org_membership) > 0:
-            return org_membership[0]
-        else:
+        if result is None:
             raise exceptions.OrganizationMembershipNotFound(membership_id)
+        return result
 
 
-class OrganizationMembershipTag(BaseTag):
+class OrganizationMembershipTag(DependentResource[OrganizationMembershipTagType]):
```
pranasziaukas, line 315: `get`, `set` and `remove` methods are missing return type annotations. Would be great to add them for consistency!

balena/models/organization.py
```diff
@@ -410,276 +237,165 @@ class OrganizationMembership:
...
-            params=params,
-            endpoint=self.settings.get("pine_endpoint"),
-            login=True,
-        )["d"]
+        org_id = organization.get(handle_or_id, {"$select": "id"})
+        org_filter = {"$filter": {"is_member_of__organization": org_id}}
+        return self.get_all(merge(org_filter, options))
 
-    def get(self, membership_id):
+    def get(self, membership_id: ResourceKey, options: AnyObject = {}):
```
pranasziaukas: Missing return type annotation.

balena/models/organization.py
```diff
@@ -410,276 +237,165 @@ class OrganizationMembership:
...
         """
         Get a single organization membership.
 
         Args:
-            membership_id (str): membership id.
+            membership_id (ResourceKey): the id (int) or an object with the unique
+            `user` & `is_member_of__organization` numeric pair of the membership
```
pranasziaukas, line 293: Missing description of the `options` argument.

balena/models/organization.py
pranasziaukas, line 1: Great work overall!

In terms of documentation, there are some minor consistency shortcomings such as:
- Not all docstrings include return statements.
- Not all docstrings include examples.
- Some docstrings are followed by trailing newlines whereas other are not.

balena/models/os.py
```diff
@@ -0,0 +1,663 @@
...
+from ..types import AnyObject
+from ..utils import compare, merge, normalize_balena_semver
+from . import application as app_module
+from .device_type import DeviceType
+
+
+class DownloadConfig(TypedDict):
+    developmentMode: NotRequired[bool]
+    appUpdatePollInterval: NotRequired[int]
+    network: NotRequired[Union[Literal["ethernet"], Literal["wifi"]]]
```
pranasziaukas: Could this be
```suggestion
    network: NotRequired[Literal["ethernet", "wifi"]]
```
?

balena/models/os.py
```diff
@@ -0,0 +1,663 @@
...
+                method="GET",
+                path=f"/device-types/v1/{slug}/images/{version}/download-size",
+            )["size"]
+        )
+
+    def get_max_satisfying_version(
+        self,
+        device_type: str,
+        version_or_range: str = "latest",
+        os_type: Optional[Literal["default", "esr"]] = None,
```
pranasziaukas, line 283: Missing docstring

balena/models/os.py
```diff
@@ -0,0 +1,663 @@
...
+    OS_UPDATE_ACTION_NAME = "resinhup"
+    RELEASE_POLICY_TAG_NAME = "release-policy"
+    ESR_NEXT_TAG_NAME = "esr-next"
+    ESR_CURRENT_TAG_NAME = "esr-current"
+    ESR_SUNSET_TAG_NAME = "esr-sunset"
+    VARIANT_TAG_NAME = "variant"
+    VERSION_TAG_NAME = "version"
+    BASED_ON_VERSION_TAG_NAME = "meta-balena-base"
+    OS_TYPES = {"default": "default", "esr": "esr"}
+    OS_VARIANTS = {"production": "prod", "development": "dev"}
```
pranasziaukas, line 204: These seem like constants but are defined as class attributes. If they don't need to change and are not intended to be overridden, they might be better placed at the module level, together with `NETWORK_WIFI`, `NETWORK_ETHERNET`, etc.

balena/models/release.py
```diff
@@ -13,70 +22,82 @@ class Release:
...
-
-    def __get_by_option(self, **options):
+        self.tags = ReleaseTag()
+
+    def __set(
+        self,
+        commit_or_id_or_raw_version: Union[str, int, ReleaseRawVersionApplicationPair],
+        body: AnyObject,
+    ) -> None:
+        id = self.get(commit_or_id_or_raw_version, {"$select": "id"})["id"]
```
pranasziaukas: Not an error, but this `id` is shadowing a built-in Python function name. According to [PEP-8](https://peps.python.org/pep-0008/#function-and-method-arguments), the way to do it is
```suggestion
        id_ = self.get(commit_or_id_or_raw_version, {"$select": "id"})["id"]
```

docs/doc2md.py
pranasziaukas, line 1: Is this some homegrown helper tool or is it standartized across several internal projects?

balena/pine.py
```diff
@@ -0,0 +1,46 @@
...
+    def __init__(self, params: Optional[Params] = None):
+        if params is None:
+            params = {}
+
+        api_url = settings.get("api_endpoint")
+        api_version = settings.get("api_version")
+
+        super().__init__({**params, "api_prefix": urljoin(api_url, api_version) + "/"})
+
+    def _request(self, method: str, url: str, body: Optional[Any] = None) -> Any:
```
pranasziaukas, line 27: Does it make sense to enforce/validate specific values for `method`?

According to the [`requests.request`](https://requests.readthedocs.io/en/latest/api/#requests.request), only `GET`, `OPTIONS`, `HEAD`, `POST`, `PUT`, `PATCH`, or `DELETE` are allowed. And our SDK may be leveraging only a subset of these.

balena/settings.py
```diff
@@ -3,6 +3,7 @@
 import os.path as Path
```
pranasziaukas, line 3: Starting with python 3.4, [`pathlib`](https://docs.python.org/3/library/pathlib.html) became (to the best of my knowledge) the de facto way of handling paths in an object-oriented way.

While you may not be able to simply replace
```suggestion
from pathlib import Path
```
without breaking anything, it may be worth it to flag this for a follow-up PR.

P.S. https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

balena/twofactor_auth.py
```diff
@@ -29,118 +23,76 @@ def is_enabled(self):
...
         Examples:
             >>> balena.twofactor_auth.is_passed()
-            True
-
         """
-
-        if not self.is_enabled():
+        try:
+            token = settings.get(TOKEN_KEY)
+            token_data = jwt.decode(token, algorithms=["HS256"], options={"verify_signature": False})
```
pranasziaukas: What are the security implications of disabling the `verify_signature` option?

My 2FA implementation knowledge is minimal but from what I can read in the [PyJWT docs](https://pyjwt.readthedocs.io/), the `verify_signature` option is enabled by default and protects against malicious parties tampering with the token. (or its signature).

balena/twofactor_auth.py
```diff
@@ -29,118 +23,76 @@ def is_enabled(self):
 
         Examples:
             >>> balena.twofactor_auth.is_enabled()
-            False
-
         """
-
         try:
-            token = self.settings.get(TOKEN_KEY)
+            token = settings.get(TOKEN_KEY)
```
pranasziaukas: `is_enabled` and `is_passed` methods share the same token decoding logic that could be refactored into a separate helper method to reduce duplication. The gains are very minimal OFC.

balena/utils.py
```diff
@@ -46,3 +55,269 @@ def compare(a, b):
...
+def is_provisioned(device: Any) -> bool:
+    return (
+        device.get("supervisor_version") is not None
+        and len(device.get("supervisor_version")) > 0
+        and device.get("last_connectivity_event") is not None
+    )
+
+
+def normalize_device_os_version(device: Any) -> Any:
+    if device.get("os_version") is not None and len(device.get("os_version")) == 0 and is_provisioned(device):
```
pranasziaukas, line 227: Looks to be the same as:
```suggestion
    if device.get("os_version") == "" and is_provisioned(device):
```

Do we really expect the `device.get("os_version")` to return an empty string?

balena/utils.py
```diff
@@ -1,22 +1,31 @@
...
-            int(value)
+            int(value)  # type: ignore
             return True
         except ValueError:
             return False
     return False
 
 
-def is_full_uuid(value):
+def is_full_uuid(value: Any) -> bool:
```
pranasziaukas, line 28: Not all strings of length 32 or 64 are valid UUIDs, see https://en.wikipedia.org/wiki/Universally_unique_identifier

A common way to check for the UUID validity seems to be something like
```python
from uuid import UUID

try:
    UUID(uuid_to_test)
except ValueError:
    return False
return True
```

pranasziaukas: Needless to say, this can also be flagged for a follow-up PR.

pranasziaukas (commented):
Once again, thank you for this massive contribution @otaviojacobi.

I have gone over the PR trying to comment on various bits and pieces. Because of the scale and scope of these breaking changes, I think some comments will end up being noise, some will be stashed for later and some might be actually useful (and might require further digging or review) -- that was exactly my intention.

Looking forward to the big release! 🚀

balena/dependent_resource.py
pranasziaukas: Thanks, I momentarily confused `Any` with `AnyObject`.

.gitignore
otaviojacobi: @pranasziaukas the lock file is not used for the customers installing the SDK, so the recomendation for libraries is to not commit it as we actually want it to alwaays run with the latest as this is what our customers will install with. This is the same behaviour we use for package-lock.json repos (such as the NOde sdk itself)

balena/balena_auth.py
otaviojacobi: @pranasziaukas  This basically matches the behaviour on the Node one, but we can discuss what would be the ideal behaviour here

balena/balena_auth.py
otaviojacobi: I think this can really be anything. Maybe could be a AnyObject but it also might be a string so not sure if Any isn't the correct typing in this very specific case

balena/balena_auth.py
otaviojacobi: I Will address the mutable parameters as a batch so it is easier to track

balena/dependent_resource.py
otaviojacobi: done

balena/dependent_resource.py
otaviojacobi: Yeah, they should all be `key` thanks!

balena/logs.py
otaviojacobi: This library (twisted) has some weird way to operate. the `transport` object there is injected at some point so type hinting goes crazy thinking that self.transport is None

balena/models/api_key.py
otaviojacobi: Flagged for follow up, but yeah, right now it just does exactly the same thing the Node one does, but we could also add validation

balena/models/application.py
otaviojacobi: Flagged for future

balena/models/billing.py
otaviojacobi: We could, but I am slightly agaisnt adding complexity to our side when we are explicitely telling (thought type hints) what values are allowed or not

balena/models/billing.py
otaviojacobi: I know... And I am actually unsure of how to properly type this. This is a requests response stream, I did not found the proper type

balena/models/credit_bundle.py
otaviojacobi: Thank you so much :)

balena/models/device.py
otaviojacobi: Flagged this and all the others. The good news is that the type hint is screaming about it, the bad one is that properly treating this is quite a bit of work, but I am up for it on a separate pr

balena/models/device.py
otaviojacobi: At the end of the day we need to keep as compactible with node sdk as possible, which uses `has` as name. And again changing this on the node one would be a big pain as we would need to update CLI, all UIs etc

balena/models/history.py
otaviojacobi: nice catch!

balena/models/organization.py
otaviojacobi: nice! fixed

balena/models/organization.py
otaviojacobi: thanks, fixed!

balena/models/os.py
otaviojacobi: yes. Much better :)

balena/models/os.py
otaviojacobi: Agree. Will keep flagged for stand alone fix

balena/models/release.py
otaviojacobi: Renamed to release_id

docs/doc2md.py
otaviojacobi: No, this is something created only for this repo and I would love to get rid of it in place of something like Sphynx

balena/pine.py
otaviojacobi: It might but this is not customer facing so as long as our own code is not doing something very weird should not be problematic, but flagged for fix

balena/twofactor_auth.py
otaviojacobi: @pranasziaukas I will be honest I don't fully understand the details on what is happening here, but AFAIK we can't really validate that this token is properly signed on the client side. The Node sdk does the same thing which it does not validate the signature

balena/twofactor_auth.py
otaviojacobi: flagged

balena/utils.py
otaviojacobi: This is mostly to keep exact the same logic as the Node one, but  we could do more in depth validation

balena/twofactor_auth.py
pranasziaukas: Thanks for clarifying!

otaviojacobi:
LGTM

PR #327, opened 2023-05-11
Allow HUP between dev OS variants

otaviojacobi:
Change-type: minor


otaviojacobi:
@thgreasi the reason why the methhod `__get_variant` is not exactly the same (typings) from the previous PR is because I just dropped the typings. The reason for that being that the typings in there consider using Semver 3.0 as parameter which are different and one of the breaking changes on that PR. All the rest, should be exactly the same. I will rebase on this once merged and I can even add the typings as a minor later on

thgreasi (approved):

Issue #337, opened 2023-07-13
Missing Functionality from balena.models.supervisor


meawoppl:
It used to be that you could initialize the `Supervisor` class with the environment variables `BALENA_SUPERVISOR_ADDRESS` and `BALENA_SUPERVISOR_API_KEY` *on a balena device* and use the class to admin the device itself from the device itself. As far as I can tell there isn't a clean/existing way to accomplish that presently.

Am I missing something? Many thanks.

otaviojacobi:
Hello @meawoppl Indeed this got past me while migrating to the new format, the current behavior will try to hit the API which will then talk to the supervisor instead of directly forwarding to the local supervisor while running the SDK locally. I will raise internally on how this can be better hanlded and come back to you

otaviojacobi:
Hello @meawoppl Could you explain what is your use case that requires using the SDK inside a device? If this is private information, could you send it to us in a private support ticket? We want to better understand the use cases here to be able to design this properly.

AmirAvonic:
We actually have the same problem and the same use case. Our devices are mostly used offline when they are used by clients, and are only attached to the Balena Cloud when a new update is needed. In the meantime, we use the Balena Python SDK in order to perform supervisor commands directly on the device, such as sending a reboot command. Is this something that will be fixed in a future release possibly?

otaviojacobi:
Hello @AmirAvonic Thanks for the clarification and explanation. This is a valid request. I will look into implementing it.

otaviojacobi:
Hello @AmirAvonic and @meawoppl  #346 has been merged, with that 14.1.0 was release and if you use SDK >= 14.1.0 the SDK running inside the device will reach the supervisor directly if you don't pass any UUID or if the UUID is equal to the same variable injected by the supervisor called BALENA_DEVICE_UUID. If the UUID is different (e.g. you are trying to do some operation from one device into another one) this will go via the API, but I believe this covers all your use cases, specially for working offline.

Please let me know if you can properly update and if you have any trouble. Note that if you are running open-balena this will require open-balena-api >= [14.1.0](https://github.com/balena-io/open-balena-api/commit/21d49241b373b2586e55ff0bebb204906dda9691)

PR #346, opened 2023-08-28
Run supervisor actions directly if on device

otaviojacobi:
Change-type: minor

### DIFF
--- a/balena/models/device.py
+++ b/balena/models/device.py
@@ -5,7 +5,9 @@ from typing import Any, Callable, List, Optional, TypedDict, Union
 from urllib.parse import urljoin
 
 from deprecated import deprecated
+import requests
 from semver.version import Version
+from json.decoder import JSONDecodeError
 
 from .. import exceptions
 from ..auth import Auth
@@ -122,6 +124,13 @@ class Device:
         self.env_var = DeviceEnvVariable(pine, self, self.__application)
         self.service_var = DeviceServiceEnvVariable(pine, self, self.__application)
         self.history = DeviceHistory(pine, settings)
+
+        self.__supervisor_address = os.environ.get("BALENA_SUPERVISOR_ADDRESS")
+        self.__supervisor_api_key = os.environ.get("BALENA_SUPERVISOR_API_KEY")
+        self.__on_device_app_id = os.environ.get("BALENA_APP_ID")
+        self.__local_device_uuid = os.environ.get("BALENA_DEVICE_UUID")
+
+        self.__on_device = all([self.__supervisor_address, self.__supervisor_api_key, self.__on_device_app_id])
 
     def __get_applied_device_config_variable_value(self, uuid_or_id: Union[str, int], name: str):
         options = {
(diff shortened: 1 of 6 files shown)


otaviojacobi:
Solves #337

balena/models/device.py
```diff
@@ -709,18 +717,46 @@ def move(
...
+        return self.__on_device and (uuid_or_id is None or uuid_or_id == self.__local_device_uuid)
+
+    def ping(self, uuid_or_id: Optional[Union[str, int]] = None) -> None:
         """
         Ping a device.
         This is useful to signal that the supervisor is alive and responding.
 
         Args:
-            uuid_or_id (Union[str, int]): device uuid (str) or id (int).
+            uuid_or_id (Optional[Union[str, int]]): device uuid (str) or id (int) or None for sdk running on device.
```
pranasziaukas: Do we want to uppercase the SDK abbreviation? I see that we're neither uppercasing ID nor UUID, but for some reason it feels odd to see sdk in lowercase. Just my 2 cents.

pranasziaukas: Do we intend to convey `None` as a technical Python keyword or none literally meaning nothing? Depending on the answer I'd choose either "None" or "none".

balena/models/device.py
```diff
@@ -709,18 +717,46 @@ def move(
...
+        req = with_supervisor_locked_error(
+            lambda: requests.request(
+                method=method, url=urljoin(self.__supervisor_address, path), json=body, params=params  # type: ignore
+            )
+        )
+
+        if req.ok:
+            try:
+                return req.json()
+            except Exception:
```
pranasziaukas: Broad exception handling can be problematic as all exceptions are treated the same way.

How about explicitly handling the most popular ones, e.g.
```suggestion
            except JSONDecodeError:
                print("Could not decode response JSON")
                return {"response": req.content.decode()}
            except Exception:
```

balena/models/device.py
```diff
@@ -809,51 +860,67 @@ def __restart_application():
...
         Examples:
             >>> balena.models.device.reboot('8f66ec7')
         """
 
-        def __reboot():
-            data = {}
-            if force:
-                data = {"force": bool(force)}
+        path = "/v1/reboot"
+        force_body = {}
```
pranasziaukas: The equivalent variable is called `force_reboot` for other methods and `force_body` for the reboot method -- should it be the other way around?

balena/models/device.py
```diff
@@ -809,51 +860,67 @@ def __restart_application():
...
         """
 
-        def __shutdown():
-            data = {}
-            if force:
-                data = {"force": bool(force)}
+        path = "/v1/shutdown"
+        force_reboot = {}
+        if force:
+            force_reboot = {"force": bool(force)}
```
pranasziaukas: Since `force` is quite a critical flag, I'd consider hard-checking the correct value type in order to absolutely minimize accidental triggers, e.g.
```python
if not isinstance(force, bool):
    raise ValueError(f"The `force` flag must be of type bool, got {type(force)} instead.")
```

WDYT?

pranasziaukas (commented):
The changes look good overall, as long as we accept the fact that the increased complexity is offloaded onto all methods, mainly in the form of `__should_run_on_device ` checks. But such split logic has to exist somewhere so there is no way around it.

balena/models/device.py
otaviojacobi: @pranasziaukas I prefer to leave "None" specifically because this is the actual check we do later (uuid_or_id is None). If you don't pass anything (none) it happens that default parameter is "None", but I would prefer to check for the actual Python keyword

balena/models/device.py
otaviojacobi: LOL. This should be "should_force" for all of them I suppose.

balena/models/device.py
otaviojacobi: great suggstion :+1:

otaviojacobi:
LGTM

PR #350, opened 2023-10-20
Add support to multipart/form-data requests

otaviojacobi:
Change-type: minor


balena/models/organization.py
```diff
@@ -26,25 +26,33 @@ def __init__(self, pine: PineClient, settings: Settings):
...
             handle (Optional[str]): The handle of the organization that will be created.
+            logo_image (Optional[io.BufferedReader]): The organization logo to be used.
 
         Returns:
             dict: organization info.
 
         Examples:
             >>> balena.models.organization.create('My Org', 'test_org')
+            >>> with open('mypath/myfile.png', 'rb') as f:
+            >>>     org = sdk.models.organization.create("my-name", None, f)
```
otaviojacobi, line 46: @pranasziaukas This (IMHO) would be the "easiest" way in Python, just `open` a file and pass it. WDYT ? On your experience usually libraries accept the open file handler or you think file path is more common used?

otaviojacobi:
LGTM

balena/models/organization.py
pranasziaukas: From my personal experience @otaviojacobi, mature libraries actually accept both. For example, [here](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html) is how the file parameter of `pandas.read_csv` is described:

> `filepath_or_buffer`: `str`, `path` object or file-like object
Any valid string path is acceptable. The string could be a URL. Valid URL schemes include http, ftp, s3, gs, and file. For file URLs, a host is expected. A local file could be: file://localhost/path/to/table.csv.
If you want to pass in a path object, pandas accepts any `os.PathLike`.
By file-like object, we refer to objects with a `read()` method, such as a file handle (e.g. via builtin `open` function) or `StringIO`.

As I recall, the same trend is true for other popular libraries as well. So it's very flexible.

That being said, I don't think everything has to be implemented at once. We have to start somewhere. Quickly thinking about the two:
- File path feels more straightforward IMHO. Users just need to provide the path to the file, and the library handles the rest. No need to know the file handling specifics in Python, etc.
- File handler arguably offers more flexibility. It allows users to open the file in different modes (e.g., read, write, binary) and handle errors, file processing, etc. before passing the file object to the library.

Just my 2 cents.

PR #355, opened 2023-12-15
Fix get_with_service_details typings

otaviojacobi:
Change-type: patch

### DIFF
--- a/balena/types/models.py
+++ b/balena/types/models.py
@@ -1,4 +1,4 @@
-from typing import Any, List, Literal, Optional, TypedDict, Union, TypeVar
+from typing import Any, List, Literal, Optional, TypedDict, Union, TypeVar, Dict
 
 __T = TypeVar("__T")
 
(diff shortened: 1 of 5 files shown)


balena/types/models.py
```diff
@@ -157,6 +158,10 @@ class TypeApplication(TypedDict):
     can_use__application_as_host: ReverseNavigationResource[ApplicationHostedOnApplication]
 
 
+class TypeApplicationWithDeviceServiceDetails(TypeApplication):
+    owns__device: List["TypeDeviceWithServices"]  # type: ignore
```
otaviojacobi, line 169: FWIW this is necessary as by default Python type hints won't allow overriding properties on `TypedDict`s

balena/types/models.py
```diff
@@ -331,6 +336,10 @@ class TypeDevice(TypedDict):
     image_install: ReverseNavigationResource[ImageInstallType]
 
 
+class TypeDeviceWithServices(TypeDevice):
+    current_services: AnyObject
```
otaviojacobi, line 340: Also, no equivalent of 
```
{ 
    [key: string]: SomeType
}
```

):

fisehara (approved):

fisehara (approved):
:+1: Thanks for improving the typings.

PR #356, opened 2024-03-25
release.get_with_with_details: Set user None when without permissions

otaviojacobi:
This was a bug on balena python sdk which would not mimic the JS sdk behaviour of setting the user as undefined when without reading permissions for it

Change-type: patch


otaviojacobi:
LGTM

pranasziaukas (approved):

PR #369, opened 2024-08-29
Api v7

otaviojacobi:
Change-type: major
See: https://balena.fibery.io/Work/Task/Update-the-balena-sdk-python-to-use-the-v7-model-2204


balena/models/os.py
```diff
@@ -493,6 +494,89 @@ def is_architecture_compatible_with(self, os_architecture: str, application_arch
 
         return True
 
+    def get_supervisor_releases_for_cpu_architecture(
+        self, cpu_architecture_slug_ord_id: Union[str, int], options: AnyObject = {}
```
myarmolinsky: ord?

balena/models/os.py
```diff
@@ -493,6 +494,89 @@ def is_architecture_compatible_with(self, os_architecture: str, application_arch
...
+                    {
+                        "$select": ["id", "raw_version", "known_issue_list"],
+                        "$filter": {
+                            "status": "success",
+                            "is_final": True,
+                            "is_invalidated": False,
+                            "semver_major": {"$gt": 0},
+                            "belongs_to__application": {
+                                "$any": {
+                                    "$alias": "a",
```
myarmolinsky, line 531: Not sure what the standard is but everywhere else I've seen we use the first letters of each word in the property for best understanding, i.e. `bta` for `belongs_to__application`, not just `a`. Not sure if this is just something Thodoris has instilled in me though rather than proper convention :P

Same for the ones below of course. Ultimately up to you of course, the goal is just for it to be understandable

tests/functional/models/test_device_os.py
```diff
@@ -38,6 +38,48 @@ def test_02_get_hup_action_type(self):
...
+            "raspberrypi4-64", {"$select": "slug", "$expand": {"is_of__cpu_architecture": {"$select": "slug"}}}
+        )
+
+        svReleases = self.balena.models.os.get_supervisor_releases_for_cpu_architecture(
+            dt["is_of__cpu_architecture"][0]["slug"]
+        )
+
+        self.assertGreater(len(svReleases), 0)
+        svRelease = svReleases[0]
+        self.assertListEqual(sorted(svRelease.keys()), sorted(["id", "raw_version", "known_issue_list"]))
```
myarmolinsky, line 57: What is the purpose of this?

balena/models/device.py
```diff
@@ -1701,35 +1701,24 @@ def set_supervisor_release(
...
-                        "$alias": "dt",
-                        "$expr": {
-                            "dt": {
-                                "slug": device_type_slug,
-                            },
-                        },
-                    },
-                },
-            },
+            "$filter": {},
```
myarmolinsky: Empty filter? Not remove?

DOCUMENTATION.md
```diff
@@ -4389,6 +4389,7 @@ The name must be a string; the optional doc argument can have any type.
     "should_be_running_on__application": Optional[List[TypeApplication]],
     "is_running_on__device": Optional[List[TypeDevice]],
     "should_be_running_on__device": Optional[List[TypeDevice]],
+    "should_operate__device": Optional[List[TypeDevice]],
```
myarmolinsky, line 4371: The commit name is `v7 model: Add the device.should_be_operated_by__release property`, is one of these a mistake?

DOCUMENTATION.md
```diff
@@ -4388,7 +4388,7 @@ The name must be a string; the optional doc argument can have any type.
     "release_image": Optional[List[ReleaseImageType]],
     "should_be_running_on__application": Optional[List[TypeApplication]],
     "is_running_on__device": Optional[List[TypeDevice]],
-    "should_be_running_on__device": Optional[List[TypeDevice]],
+    "is_pinned_to__device": Optional[List[TypeDevice]],
```
myarmolinsky, line 4370: The commit name is `v7 model: Replace the device.should_be_running__release with is_pinned_on__release`, should this be in a different commit dedicated to adding `is_pinned_to__device`?

DOCUMENTATION.md
```diff
@@ -4196,7 +4196,7 @@ The name must be a string; the optional doc argument can have any type.
     "start_timestamp": str,
     "end_timestamp": str,
     "push_timestamp": str,
-    "image_size": int,
+    "image_size": str,
```
myarmolinsky, line 4178: Just to make sure as I'm unfamiliar with python:
This is actually a big int, is `str` the correct type for that?

thgreasi: pine converts big int values to a string (which is also covered by the OData spec), so that they can be serialized in JSON (since json, just like plain js doesn't support too large numeric values.

DOCUMENTATION.md
thgreasi: I guess we could do a separate commit strictly speaking.
In the js SDK I chose to update each term in a separate commit, so I grouped both the normal fact type (`device.is_pinned_on__release`) and its new synonymous form (`release.is_pinned_to__device`) in the same commit.

DOCUMENTATION.md
thgreasi: +1, it indeed seems that we are missing the `device.should_be_operated_by__release`

balena/types/models.py
```diff
@@ -335,7 +336,7 @@ class TypeDevice(TypedDict):
     is_running__release: OptionalNavigationResource["ReleaseType"]
     should_be_running__release: OptionalNavigationResource["ReleaseType"]
     is_managed_by__service_instance: OptionalNavigationResource[ServiceInstanceType]
-    should_be_managed_by__supervisor_release: OptionalNavigationResource[SupervisorReleaseType]
+    should_be_managed_by__release: OptionalNavigationResource[SupervisorReleaseType]
```
thgreasi: Let's also add the synonymous form `release.should_manage__device` in the same PR so that it includes both the normal fact type as well as its synonymous form.

thgreasi:
```suggestion
    should_be_managed_by__release: OptionalNavigationResource[ReleaseType]
```

balena/models/device.py
thgreasi, line 1: Heads-up reminder to also drop the `SupervisorReleaseType`, since the commit is callled `v7 model: Drop the supervisor_release resource` :)

balena/models/os.py
otaviojacobi: Just keeping parity with nodejs sdk query which I use always as source of truth: https://github.com/balena-io/balena-sdk/blob/88070df6f13c21db0c6f83cc1683f46a1a8f4660/src/models/os.ts#L1009

I think the reasoning behind here is that this query is already huge and the expanded resource is just an `a`application

balena/models/os.py
otaviojacobi: Nice find, it was a typo, thank you!

tests/functional/models/test_device_os.py
otaviojacobi: In short, the new function `get_supervisor_releases_for_cpu_architecture` should alawys return  these 3 keys, no matter the value (e.g. it can be null but not undefined in javascript world). So the test basically ensures that

balena/models/device.py
otaviojacobi: So the next two lines always ensure this would be fill in. However, I found a better python syntax for it and updated the PR.

DOCUMENTATION.md
otaviojacobi: I legit missed this commit. Thanks for noticing, added!

DOCUMENTATION.md
otaviojacobi: I added both in the same commit as a pattern. In the node one we sometimes did it together and sometimes spplited. I decided to make a single commit entry for adding both the regular form and the synonymous form, I think it makes more clear to read and creates less unnecessary commit entries.

DOCUMENTATION.md
otaviojacobi: Altough python has its own bigint class I would prefer (at least for now) to keep things as a string as it matches the typing on the api. I think it is a nice functionality for the SDK to automatically convert these strings into python bigint but needs to be done across the entire sdk rather than in a single point.

myarmolinsky (approved):

PR #371, opened 2024-10-14
do not operate on empty or ambiguous device and release UUIDs 

otaviojacobi:
Change-type: patch

### DIFF
--- a/balena/models/device.py
+++ b/balena/models/device.py
@@ -174,6 +174,19 @@ class Device:
                     }
                 )
             else:
+                if len(uuid_or_id_or_ids) < 6:
+                    raise exceptions.InvalidParameter("UUID must have at least 6 characeters", None)
+
+                affected = self.__pine.get({
+                    "resource": "device",
+                    "options": {
+                        "$top": 2,
+                        "$select": "id",
+                        "$filter": {"uuid": {"$startswith": uuid_or_id_or_ids}}}
+                    })
+                if len(affected) > 1:
+                    raise exceptions.AmbiguousDevice(uuid_or_id_or_ids)
+
                 fn(
                     {
                         "resource": "device",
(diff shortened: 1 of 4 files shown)


balena/models/device.py
```diff
@@ -174,6 +174,19 @@ def __set(
...
             else:
+                if len(uuid_or_id_or_ids) < 6:
+                    raise exceptions.AmbiguousDevice(uuid_or_id_or_ids)
+
+                affected = self.__pine.get({
+                    "resource": "device",
+                    "options": {
+                        "$count": {
+                           "$filter": {"uuid": {"$startswith": uuid_or_id_or_ids}}}
+                        }
```
thgreasi: This should be marginally faster :)
```suggestion
                        "$top": 2,
                        "$select": "id",
                        "$filter": {"uuid": {"$startswith": uuid_or_id_or_ids}}}
```

balena/models/device.py
```diff
@@ -174,6 +174,19 @@ def __set(
                     }
                 )
             else:
+                if len(uuid_or_id_or_ids) < 6:
+                    raise exceptions.AmbiguousDevice(uuid_or_id_or_ids)
```
thgreasi: I guess an InvalidParameter error might have been better, explaining that they need to provide at least 6 chars

thgreasi:
`release.get` should probably also throw if you pass an empty string.
We could then on the next major bump that to 5 chars min, since that's the smallest possible semver.

thgreasi:
Oh and `device.get` could similarly fail w/  an InvalidParameter error when providing an empty string

balena/models/release.py
```diff
@@ -57,6 +57,9 @@ def get(
         if commit_or_id_or_raw_version is None:
             raise exceptions.ReleaseNotFound(commit_or_id_or_raw_version)
 
+        if commit_or_id_or_raw_version == '':
+            raise exceptions.InvalidParameter("UUID can not be empty", None)
```
thgreasi:
```suggestion
            raise exceptions.InvalidParameter("commit_or_id_or_raw_version can not be empty", None)
```

thgreasi (approved):
Converted to draft so that you have the chance to fix the invalid parameter error and not get blocked by my pending review

PR #378, opened 2025-01-29
Add option for detached HUP on start_os_update

jaomaloy:
This will call the v2 actions endpoint for resinhup which runs a detached version of HUP that increases HUP reliability on slow networks but will offer
no status updates such as in_progress.

Change-type: minor

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -1976,6 +1976,9 @@ Start an OS update on a device.
         Unsupported (unpublished) version will result in rejection.
         The version **must** be the exact version number, a "prod" variant
         and greater than the one running on the device.
+    run_detached (Optional[bool]): run the update in detached mode.
+        Default behaviour is run_detached=False but is DEPRECATED and will be
+        removed in a future release. Use run_detached=True for more reliable updates.
 
 #### Returns:
     HUPStatusResponse: action response.
@@ -1984,6 +1987,7 @@ Start an OS update on a device.
 ```python
 >>> balena.models.device.start_os_update('b6070f4', '2.29.2+rev1.prod')
 >>> balena.models.device.start_os_update('b6070f4', '2.89.0+rev1')
+>>> balena.models.device.start_os_update('b6070f4', '2.89.0+rev1', run_detached=True)
 ```
 
 <a name="device.start_service"></a>
--- a/balena/models/device.py
+++ b/balena/models/device.py
@@ -1727,7 +1727,13 @@ class Device:
             }
         )
 
-    def start_os_update(self, uuid_or_id: Union[str, int], target_os_version: str) -> HUPStatusResponse:
+    def start_os_update(
+        self,
+        uuid_or_id: Union[str, int],
+        target_os_version: str,
+        *,  # Force keyword arguments after this point
+        run_detached: Optional[bool] = False
+    ) -> HUPStatusResponse:
         """
         Start an OS update on a device.
 
@@ -1737,6 +1743,9 @@ class Device:
                 Unsupported (unpublished) version will result in rejection.
                 The version **must** be the exact version number, a "prod" variant
                 and greater than the one running on the device.
+            run_detached (Optional[bool]): run the update in detached mode.
+                Default behaviour is run_detached=False but is DEPRECATED and will be
+                removed in a future release. Use run_detached=True for more reliable updates.
 
         Returns:
             HUPStatusResponse: action response.
@@ -1744,6 +1753,7 @@ class Device:
         Examples:
             >>> balena.models.device.start_os_update('b6070f4', '2.29.2+rev1.prod')
             >>> balena.models.device.start_os_update('b6070f4', '2.89.0+rev1')
+            >>> balena.models.device.start_os_update('b6070f4', '2.89.0+rev1', run_detached=True)
         """
 
         if target_os_version is None or uuid_or_id is None:
@@ -1766,7 +1776,9 @@ class Device:
         data = {"parameters": {"target_version": target_os_version}}
 
         url_base = self.__config.get_all()["deviceUrlsBase"]
-        action_api_version = self.__settings.get("device_actions_endpoint_version")
+        if not isinstance(run_detached, bool):
+            raise ValueError(f"run_detached must be True or False, got {type(run_detached)}: {run_detached}")
+        action_api_version = "v2" if run_detached is True else self.__settings.get("device_actions_endpoint_version")
 
         return request(
             method="POST",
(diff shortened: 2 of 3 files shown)


DOCUMENTATION.md
```diff
@@ -1976,6 +1976,9 @@ Start an OS update on a device.
         Unsupported (unpublished) version will result in rejection.
         The version **must** be the exact version number, a "prod" variant
         and greater than the one running on the device.
+    run_detached (Optional[bool]): run the update in detached mode.
+        Default behaviour is run_detached=False but is DEPRECATED and will be
+        removed in a future release. Use run_detached=True.
```
kb2ma: Include the rationale, as we do in the Node SDK, with "Use run_detached=True for more reliable updates"

balena/models/device.py
```diff
@@ -1737,15 +1743,23 @@ def start_os_update(self, uuid_or_id: Union[str, int], target_os_version: str) -
                 Unsupported (unpublished) version will result in rejection.
                 The version **must** be the exact version number, a "prod" variant
                 and greater than the one running on the device.
+            run_detached (Optional[bool]): run the update in detached mode.
+                Default behaviour is run_detached=False but is DEPRECATED and will be
+                removed in a future release. Use run_detached=True.
```
kb2ma: Include the rationale, as we do in the Node SDK, with "Use run_detached=True for more reliable updates"

balena/models/device.py
```diff
@@ -1737,15 +1743,23 @@ def start_os_update(self, uuid_or_id: Union[str, int], target_os_version: str) -
...
         Returns:
             HUPStatusResponse: action response.
 
         Examples:
             >>> balena.models.device.start_os_update('b6070f4', '2.29.2+rev1.prod')
             >>> balena.models.device.start_os_update('b6070f4', '2.89.0+rev1')
+            >>> balena.models.device.start_os_update('b6070f4', '2.89.0+rev1', run_detached=True)
         """
 
+        # Set default value for run_detached if None
```
kb2ma: Simpler and clearer to set a default value for the argument in the function definition.

balena/models/device.py
```diff
@@ -1776,6 +1790,7 @@ def start_os_update(self, uuid_or_id: Union[str, int], target_os_version: str) -
             endpoint=f"https://actions.{url_base}/{action_api_version}/",
         )
 
+    @deprecated("This will be removed in a future major release")
```
kb2ma: Include the rationale, as we do in the Node SDK , with "This will no longer return a useful status for runDetached=true updates."

kb2ma (changes requested):
First pass review. Have not tested operation yet.

balena/models/device.py
```diff
@@ -1766,7 +1776,7 @@ def start_os_update(self, uuid_or_id: Union[str, int], target_os_version: str) -
         data = {"parameters": {"target_version": target_os_version}}
 
         url_base = self.__config.get_all()["deviceUrlsBase"]
-        action_api_version = self.__settings.get("device_actions_endpoint_version")
+        action_api_version = "v2" if run_detached is True else self.__settings.get("device_actions_endpoint_version")
```
otaviojacobi, line 1781:
```suggestion
        action_api_version = "v2" if run_detached else self.__settings.get("device_actions_endpoint_version")
```

otaviojacobi: @kb2ma @jaomaloy Are we using `v2` actions endpoint only on for HuP or should the entire SDK move to the new actions endpoint?

jaomaloy: v2 is only for detached HUP at this point @otaviojacobi

kb2ma:
Tested HUP against live device with and without `runDetached=True`, and it worked as expected.

Added a commit with some simple input tests on `start_os_update()`.

@jaomaloy if you can squash in my commit and fix the issue flagged by otaviojacoby, I'm happy to approve. Of course let me know if you have any concerns about the test I added.

balena/models/device.py
kb2ma: Good catch on the "is True", @otaviojacobi . If the user happens to use runDetached="True", that still works with your change since the test now is for Truthiness. I'm a little surprised use of a string isn't type checked though since it's typed as `Optional[bool]` .

otaviojacobi: Yes, the key thing here is to (similar to typescript) always think of the type system as best effort. There is only so much you can do on a dynamically typed languages (js/python) to avoid these mistakes. There is the argument of if someone does `run_detatched="False"` it will still be truthy, however considering it defaults to `False` I think someone messing with this parameter probably wants to run detatched

otaviojacobi: We can also always validate that it is either  `True` `False` or `None` and throw it otherwise so that we avoid all these sort of issues. I would defer if you want to go this route to you

kb2ma: I actually like the strict checking best to remove the ambiguity.

jaomaloy: Makes sense. Added below.

```
if not isinstance(run_detached, bool) and run_detached is not None:
            raise ValueError(f"run_detached must be True, False, or None, got {type(run_detached)}: {run_detached}")
action_api_version = "v2" if run_detached is True else self.__settings.get("device_actions_endpoint_version")
```

kb2ma: I don't see why we need to allow the value to be None. We have  a default value of False. How about:

```
if not isinstance(run_detached, bool):
    raise ValueError(f"run_detached must be True or False, got {type(run_detached)}: {run_detached}")
action_api_version = "v2" if run_detached else self.__settings.get("device_actions_endpoint_version")
```

kb2ma:
Testing with 7ab93c0 below. Works as expected. The _status_ value on `start_os_update()` shows running the appropriate HUP flavor.

```
<stdin>:1: DeprecationWarning: Call to deprecated method get_os_update_status. (
        This will be removed in a future major release. This will no longer return a
        useful status for runDetached=true updates.
        )
{'status': 'idle', 'action': 'resinhup', 'lastRun': None}

>>> balena.models.device.start_os_update('c575748107105508f1eb8dc69b6e1e19', '6.1.21', run_detached=None)
  File "<stdin>", line 1, in <module>
  File "/home/kbee/dev/python-sdk/repo/balena/models/device.py", line 1780, in start_os_update
    raise ValueError(f"run_detached must be True or False, got {type(run_detached)}: {run_detached}")
ValueError: run_detached must be True or False, got <class 'NoneType'>: None

>>> balena.models.device.start_os_update('c575748107105508f1eb8dc69b6e1e19', '6.1.21', run_detached=True)
{'status': 'triggered', 'lastRun': 1739293507277, 'parameters': {'target_version': '6.1.21'}, 'action': 'resinhup'}

>>> balena.models.device.start_os_update('c575748107105508f1eb8dc69b6e1e19', '6.1.21+rev1')
{'status': 'in_progress', 'lastRun': 1739293852176, 'parameters': {'target_version': '6.1.21+rev1'}, 'action': 'resinhup'}

>>> balena.models.device.start_os_update('c575748107105508f1eb8dc69b6e1e19', '6.1.24', run_detached=False)
{'status': 'in_progress', 'lastRun': 1739294909010, 'parameters': {'target_version': '6.1.24'}, 'action': 'resinhup'}
```

kb2ma (approved):
Latest update looks good.

PR #379, opened 2025-02-05
Fix broken tar tests URLs to match same from the Node SDK

otaviojacobi:
Change-type: patch

### DIFF
--- a/tests/functional/models/test_release.py
+++ b/tests/functional/models/test_release.py
@@ -16,7 +16,7 @@ class TestRelease(unittest.TestCase):
             "ServiceTestApp", "raspberry-pi2", cls.helper.default_organization["id"]
         )
         cls.mc_app = cls.helper.create_multicontainer_app()
-        cls.TEST_SOURCE_URL = "https://github.com/balena-io-examples/balena-node-hello-world/archive/v1.0.0.tar.gz"
+        cls.TEST_SOURCE_URL = "https://codeload.github.com/balena-io-examples/balena-nodejs-hello-world/tar.gz/refs/tags/v1.0.0"
         cls.TEST_SOURCE_CONTAINER_COUNT = 1
 
         # releases can be queried by id, commit or a dict containing application and raw_version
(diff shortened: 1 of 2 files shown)


tests/functional/models/test_release.py
```diff
@@ -67,16 +67,17 @@ def test_03_create_from_url(self):
         with self.assertRaises(self.helper.balena_exceptions.ApplicationNotFound):
             self.balena.models.release.create_from_url(123, url=self.TEST_SOURCE_URL)
 
-        with self.assertRaises(self.helper.balena_exceptions.BuilderRequestError) as cm:
-            self.balena.models.release.create_from_url(
-                self.empty_app["id"],
-                url="https://github.com/balena-io-projects/simple-server-node/archive/v0.0.0.tar.gz",
-            )
-        self.assertIn("Failed to fetch tarball from passed URL", cm.exception.message)
```
otaviojacobi, line 75: Here is my understanding from what happened and why the testing was the way it was: when I refactored this sdk a few years back and tried to setup it up similar to the node-sdk in this test I tested for the `actual` behavior (which would be to fail to fetch the tarball because the status code would be a 302 redirect from the balena-io-projects org to a new one) rather than the correct behavior which is to actually create a release

The new test should be much better and similar to the one in the node-sdk now.

tests/functional/test_logs.py
```diff
@@ -35,7 +35,7 @@ def setUp(self):
         registration_info = self.balena.models.device.register(self.app["id"], self.uuid)
         self.device_api_key = registration_info["api_key"]
 
-    def __collect_logs(self, timeout: int = 5, count=None):
+    def __collect_logs(self, timeout: int = 30, count=None):
```
otaviojacobi: Because our tests run agaisnt production I think these might have slowed a bit. I will check to see if we have enabled loki read logs which could potentially have slwoed things a bit.

kb2ma:
Tests passed for me locally on commit [e388a49](https://github.com/balena-io/balena-sdk-python/pull/379/commits/e388a49a60969039297b0e6a9f771a00ecfa38b0).

tests/functional/test_logs.py
```diff
@@ -5,6 +5,11 @@
 from tests.helper import TestHelper
 import time
 
+WAIT_FOR_FIRSTLOGS_TIMEOUT_S = 5
```
kb2ma, line 22: Please add some comments here on why these different timeouts are required and how you chose the values. This will help if we ever need to adjust in the future.

kb2ma (changes requested):
I generally understand the approach here as a way to be more precise on timeouts. The bottom line is that tests are passing now. See the one request for some rationale for future readers.

tests/functional/test_logs.py
otaviojacobi: Agree!

kb2ma (approved):
Really nice explanation on the timeouts, thanks.

PR #381, opened 2025-02-08
Validate target OS update version is not same as current

kb2ma:
The balena-hup-action-utils package [validates](https://github.com/balena-io-modules/balena-hup-action-utils/blob/3cc85eece148652448445851a90d5620b4f7166b/lib/index.ts#L103) this condition for the Node SDK. The Python SDK must do the same. The docs [say](https://docs.balena.io/reference/sdk/python-sdk/latest/#function-start_os_updateuuid_or_id-target_os_version-%E2%87%92-codehupstatusresponsecode) it does. ;-)

This PR is a WIP. Need to resolve how build versions compare. Presently test fails for x.y.z+rev1 vs. x.y.z+rev2 -- they compare as equal.


kb2ma:
Superseded by #393

PR #382, opened 2025-02-12
Enhance documentation for detached_mode OS update

kb2ma:
Improves updates in v15.1.0 for `start_os_update() `and `get_os_update_status()`.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -1539,6 +1539,7 @@ Get device name by device uuid.
 <a name="device.get_os_update_status"></a>
 ### Function: get_os_update_status(uuid_or_id) ⇒ <code>HUPStatusResponse</code>
 
+***Deprecated***
 Get the OS update status of a device.
 
 #### Args:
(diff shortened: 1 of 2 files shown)


DOCUMENTATION.md
```diff
@@ -1539,6 +1539,7 @@ Get device name by device uuid.
 <a name="device.get_os_update_status"></a>
 ### Function: get_os_update_status(uuid_or_id) ⇒ <code>HUPStatusResponse</code>
 
+***Deprecated***
```
jaomaloy, line 1542: Maybe we could add

"If using run_detached option, monitor progress with device.get() --
status, provisioning_state and provisioning_progress entries."

here too?

jaomaloy (approved):

DOCUMENTATION.md
kb2ma: I see what you're driving at, but `get_os_update_status()` still is valid if not using the run_detached option. This function will go away with the next major, so will become moot.

PR #384, opened 2025-02-18
tests: Increase logs timeouts

otaviojacobi:
Change-type: patch


kb2ma (approved):
This is a trivial change that takes advantage of the new configurable timeouts for this test.

PR #387, opened 2025-05-22
Do not throw in case of failure to refresh token

otaviojacobi:
Change-type: patch


JSReds:
LGTM, just wondering if we should have approximatelly the same approach as we have node-dek https://github.com/balena-io-modules/balena-auth/blob/ed8a48b01e8815e17f803a3e9b426da27fdf9b5a/src/auth.ts#L217

JSReds (approved):

PR #389, opened 2026-01-23
No longer support using device short uuid

otaviojacobi:
Change-type: major

### DIFF
--- a/balena/models/device.py
+++ b/balena/models/device.py
@@ -164,40 +164,20 @@ class Device:
             fn = self.__pine.patch
 
         if isinstance(uuid_or_id_or_ids, (int, str)):
-            is_potentially_full_uuid = is_full_uuid(uuid_or_id_or_ids)
-            if is_potentially_full_uuid or is_id(uuid_or_id_or_ids):
-                fn(
-                    {
-                        "resource": "device",
-                        "id": uuid_or_id_or_ids if is_id(uuid_or_id_or_ids) else {"uuid": uuid_or_id_or_ids},
-                        "body": body,
-                    }
-                )
+            if is_id(uuid_or_id_or_ids):
+                resource_id = uuid_or_id_or_ids
+            elif is_full_uuid(uuid_or_id_or_ids):
+                resource_id = {"uuid": uuid_or_id_or_ids}
             else:
-                if len(uuid_or_id_or_ids) < 6:
-                    raise exceptions.InvalidParameter("UUID must have at least 6 characeters", None)
+                raise exceptions.InvalidParameter("uuid_or_id_or_ids", uuid_or_id_or_ids)
 
-                affected = self.__pine.get(
-                    {
-                        "resource": "device",
-                        "options": {
-                            "$top": 2,
-                            "$select": "id",
-                            "$filter": {"uuid": {"$startswith": uuid_or_id_or_ids}},
-                        },
-                    }
-                )
-                if len(affected) > 1:
-                    raise exceptions.AmbiguousDevice(uuid_or_id_or_ids)
-
-                fn(
-                    {
-                        "resource": "device",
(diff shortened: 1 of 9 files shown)


tests/functional/models/test_api_key.py
```diff
@@ -86,7 +86,7 @@ def test_06_get_provisioning_api_keys_by_application(self):
         self.assertIn("provisionTestKey", provisioning_keys_names)
 
     def test_07_get_device_api_keys_by_device_for_non_existing(self):
-        with self.assertRaises(self.helper.balena_exceptions.DeviceNotFound):
+        with self.assertRaises(self.helper.balena_exceptions.InvalidParameter):
             self.balena.models.api_key.get_device_api_keys_by_device("nonexistentuuid")
```
thgreasi, line 90: I think we should have one test case that throws a DeviceNotFound error when using a random uuid, and ideally (but optionally since that's tested elsewhere as well) we can also have one to confirm that a short uuid throws an InvalidParameter error.

balena/models/device.py
```diff
@@ -1757,10 +1725,10 @@ def start_os_update(
...
 
         Examples:
-            >>> balena.models.device.start_os_update('b6070f4', '2.29.2+rev1.prod')
-            >>> balena.models.device.start_os_update('b6070f4', '2.89.0+rev1')
-            >>> balena.models.device.start_os_update('b6070f4', '2.89.0+rev1', run_detached=True)
-        """
+            >>> balena.models.device.start_os_update('b6070f4fea5a4f11b4d05c1f1c3b4e72', '2.29.2+rev1.prod')
+            >>> balena.models.device.start_os_update('b6070f4fea5a4f11b4d05c1f1c3b4e72', '2.89.0+rev1')
+            >>> balena.models.device.start_os_update('b6070f4fea5a4f11b4d05c1f1c3b4e72', '2.89.0+rev1', run_detached=True)
+        """  # noqa: E501
```
thgreasi, line 1731: ???
```suggestion
        """
```

tests/functional/models/test_release.py
```diff
@@ -153,9 +153,6 @@ def test_06_get_release_by_resources(self):
         release = self.balena.models.release.get(self.mc_app["current_release"]["commit"])
         self.__expect_release_to_match_on_get(release, match)
 
-        release = self.balena.models.release.get(self.mc_app["current_release"]["commit"][0:7])
-        self.__expect_release_to_match_on_get(release, match)
```
thgreasi, line 157: Can we re-add this test back and have it test that it throws the correct error / doesn't find the release?

tests/functional/models/test_release.py
```diff
@@ -289,59 +280,6 @@ def test_12_get_latest_by_application(self):
...
-        with self.assertRaises(self.helper.balena_exceptions.AmbiguousRelease):
-            self.balena.models.release.get("feb23612")
-
-        release = self.balena.models.release.get("feb2361230dc40dba6dca9a18f2c19dc8f2c19dc")
-
-        self.assertEqual(release["commit"], "feb2361230dc40dba6dca9a18f2c19dc8f2c19dc")
-        self.assertEqual(release["status"], "success")
-        self.assertEqual(release["source"], "cloud")
-
-        with self.assertRaises(self.helper.balena_exceptions.AmbiguousRelease):
```
thgreasi, line 334: Would be nice to have a test that confirms that a AmbiguousRelease error is thrown when two releases across different apps are found, or two releases within the same app (one successful & one failed)

balena/models/device.py
otaviojacobi: It is necessary @thgreasi as the doc line above (run_detatcher etc) goes above the max length line size

thgreasi: til

thgreasi (approved):
Let's just add a DeviceNotFound test as well

PR #390, opened 2026-02-02
Fix linter

otaviojacobi:
Change-type: patch

### DIFF
--- a/tests/functional/models/test_application.py
+++ b/tests/functional/models/test_application.py
@@ -236,7 +236,8 @@ class TestApplication(unittest.TestCase):
 
     def test_31_membership_create(self):
         app = TestApplication.app
-        membership = self.balena.models.application.membership.create(app["id"], "device_tester1")
+        member_username = self.helper.member_credentials["member_username"]
+        membership = self.balena.models.application.membership.create(app["id"], member_username)
         TestApplication.membership = membership
         self.assertEqual(membership["is_member_of__application"]["__id"], app["id"])
 
--- a/tests/functional/models/test_application.py
+++ b/tests/functional/models/test_application.py
@@ -68,9 +68,7 @@ class TestApplication(unittest.TestCase):
             self.balena.models.application.get_all_by_organization(self.org_handle)[0]["app_name"], "FooBar"
         )
 
-        self.assertEqual(
-            self.balena.models.application.get_all_by_organization(self.org_id)[0]["app_name"], "FooBar"
-        )
+        self.assertEqual(self.balena.models.application.get_all_by_organization(self.org_id)[0]["app_name"], "FooBar")
 
     def test_07_has(self):
         self.assertFalse(self.balena.models.application.has("AppNotExist"))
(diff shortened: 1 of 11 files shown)


tests/functional/models/test_application.py
```diff
@@ -238,7 +236,8 @@ def test_30_membership_get_all_by_application_empty(self):
 
     def test_31_membership_create(self):
         app = TestApplication.app
-        membership = self.balena.models.application.membership.create(app["id"], "device_tester1")
```
thgreasi, line 241: `"device_tester1" ❗

tests/helper.py
```diff
@@ -73,8 +90,14 @@ def load_env(cls):
...
                 cls.credentials["api_endpoint"] = os.environ.get("TEST_API_ENDPOINT")
-            except Exception:
-                raise Exception("Mandatory env keys missing!")
+
+                cls.member_credentials = {
+                    "member_email": os.environ["TEST_MEMBER_EMAIL"],
+                    "member_username": os.environ["TEST_MEMBER_USER_ID"],
+                    "member_password": os.environ["TEST_MEMBER_PASSWORD"],
+                }
+            except KeyError as e:
```
thgreasi, line 99: Where is the KeyError thrown from ?

tests/helper.py
```diff
@@ -84,10 +107,11 @@ def wipe_application(self):
 
     def wipe_organization(self):
         """
-        Wipe all user's orgs
+        Wipe all user's orgs except the personal org
```
thgreasi, line 110: How about having a well known org name prefix and deleting only those orgs like we do in the node sdk?
That way you will not have to say the cursed word "personal" anywhere.
https://github.com/balena-io/balena-sdk/blob/master/tests/integration/setup.ts#L295-L314

tests/functional/models/test_tag.py
```diff
@@ -98,7 +98,7 @@ def test_set(self, test_runner, resource_id):
         test_runner.assertIn(
             (
                 f"It is necessary that each {self.associated_resource_name} tag has a "
-                "tag key that has a Length (Type) that is greater than 0."
```
thgreasi, line 101: Is this `.` no longer necessary or thrown by the API?

thgreasi (approved):

tests/functional/models/test_tag.py
otaviojacobi: @thgreasi the new constraint is `that is greater than 0 and smaller than 100/164(depending on the tag type afaict). - this makes the test accept all cases and simply tests we are returning a fair enough error message

tests/helper.py
otaviojacobi: This is just a slightly improvement over the previous Exception that would catch all errors as env keys missing. If any variable is missing `os.environ["TEST_MEMBER_EMAIL"]` or any other access will throw an specific KeyError

tests/functional/models/test_application.py
otaviojacobi: Yeah.... but it is not problematic as in the sense no fleet here has access to anything anyway

tests/helper.py
otaviojacobi: I would be fine in a follow up on review/clean up week :) added to my todo aswell

PR #391, opened 2026-02-16
Block OS upgrades from current OS < 2.14.0 and target OS < 2.16.0

kb2ma:
Update the minimum target OS to v2.16.0 for `start_os_update()`. Also add a minimum for currently running OS to v2.14.0. This value essentially means current OS is running from a Docker image. These values are required to support v4 of the balenahup script, which removes support for older OSes. See balena-os/balenahup#422.

Also a couple of other related simple but major changes:
* Change the default for `start_os_update()` to use detached mode for more reliable updates
* Rename `set_supervisor_release()` to `pin_to_supervisor_release()`

See: https://balena.fibery.io/Work/Project/2183

### DIFF
--- a/balena/hup.py
+++ b/balena/hup.py
@@ -4,7 +4,10 @@ from semver.version import Version
 
 from . import exceptions
 
-MIN_TARGET_VERSION = "2.2.0+rev1"
+# The HUP process itself as well as policies set minimum bounds on current/target
+# versions that it can update. These variables allow enforcement of these minimums.
+MIN_CURRENT_VERSION = "2.14.0+rev1"
+MIN_TARGET_VERSION = "2.16.0+rev1"
 
 
 def __get_variant(ver: Version) -> Optional[Literal["dev", "prod"]]:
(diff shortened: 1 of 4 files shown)


balena/hup.py
```diff
@@ -45,8 +48,10 @@ def get_hup_action_type(device_type: str, current_version: Optional[str], target
     if Version.parse(target_version).compare(current_version) < 0:
         raise exceptions.OsUpdateError("OS downgrades are not allowed")
 
-    # For 1.x -> 2.x or 2.x to 2.x only
-    if parsed_target_ver.major > 1 and Version.parse(target_version).compare(MIN_TARGET_VERSION) < 0:
+    if Version.parse(current_version).compare(MIN_CURRENT_VERSION) < 0:
+        raise exceptions.OsUpdateError("Current balenaOS version must be greater than {0}".format(MIN_CURRENT_VERSION))
```
thgreasi: Shouldn't the error say "greater than or equal to" in both cases?

thgreasi (approved):

balena/hup.py
kb2ma: Good catch; fixed.

balena/models/device.py
```diff
@@ -1655,7 +1655,7 @@ def track_application_release(self, uuid_or_id_or_ids: Union[str, int, List[int]
         self.__set(uuid_or_id_or_ids, {"is_pinned_on__release": None})
 
     # TODO: enable device batching
-    def set_supervisor_release(
+    def pin_to_supervisor_release(
```
thgreasi, line 1658: You also need to regenerate the docs afaict.
The README says to run v
```
python docs_generator.py > DOCUMENTATION.md
```
other than that lgtm

kb2ma: Also fixed up a couple of doc comment typos found during review.

PR #392, opened 2026-03-18
Add pin_to_os_release() for offline updates

kb2ma:
Adds the ability to pin a device to an OS version with `pin_to_os_release()`. Similar to app updates, the OS can be pinned while the device is offline. When the device comes online, we expect the cloud to push the update to the device.

A device may be pinned back to the current OS version. In contrast, `start_os_update()` now requires that the target OS version is not the current OS version.

Depends on #393.

Change-type: minor

### DIFF
--- a/balena/models/device.py
+++ b/balena/models/device.py
@@ -1,12 +1,14 @@
 import binascii
 import datetime
 import os
+import re
 from typing import Any, Callable, List, Optional, TypedDict, Union, cast
 from urllib.parse import urljoin
 
 from deprecated import deprecated
 import requests
 from semver.version import Version
+from semver import compare as semver_compare
 from json.decoder import JSONDecodeError
 
 from .. import exceptions
@@ -211,12 +213,15 @@ class Device:
         if device["os_variant"] != "dev":
             raise exceptions.LocalModeError(Message.DEVICE_OS_TYPE_NOT_SUPPORT_LOCAL_MODE)
 
-    def __check_os_update_target(self, device_info: TypeDevice, target_os_version: str):
+    def __check_os_update_target(self, device_info: TypeDevice, target_os_version: str, mode: str):
+        # Mode must be "start" (start_os_update) to validate device is online.
+        # Otherwise, mode must be "pin" (pin_to_os_release) to relax version comparison
+        # as commented below.
         if "uuid" not in device_info or not device_info["uuid"]:
             raise exceptions.OsUpdateError("The uuid of the device is not available")
 
         uuid = device_info["uuid"]
-        if "is_online" not in device_info or not device_info["is_online"]:
+        if mode == "start" and ("is_online" not in device_info or not device_info["is_online"]):
             raise exceptions.OsUpdateError(f"The device is offline: {uuid}")
 
         if "os_version" not in device_info or not device_info["os_version"]:
@@ -227,6 +232,28 @@ class Device:
 
         if "os_variant" not in device_info:
             raise exceptions.OsUpdateError(f"The os variant of the device is not available: {uuid}")
+
+        # Allow pinning back to current version. This case is specific to pinning because it
+        # is asynchronous, while starting is synchronous.
+        # If versions differ, continue to follow the logic in get_hup_action_type() below.
+        # The code here does not validate minimum OS version. The goal is to disable an earlier
+        # pin to a forward version; so the device already is running the "target" version.
+        if mode == "pin":
+            try:
+                # Must normalize first, since in format like "balenaOS 6.5.19".
+                parsed_current_ver = normalize_balena_semver(device_info["os_version"])
+                # Collect target version and variant.
+                res = re.match(r"^(.+?)(?:\.(dev|prod))?$", target_os_version)
+                # Ensure version and variant match.
+                if (
+                    res.group(1)
+                    and (not res.group(2) or device_info["os_variant"] == res.group(2))
+                    and semver_compare(parsed_current_ver, res.group(1)) == 0
+                ):
+                    return
+            except Exception:
+                # Will be handled below in get_hup_action_type()
+                pass
 
         current_os_version = get_device_os_semver_with_variant(device_info["os_version"], device_info["os_variant"])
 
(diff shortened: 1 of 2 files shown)


balena/models/device.py
```diff
@@ -228,6 +233,21 @@ def __check_os_update_target(self, device_info: TypeDevice, target_os_version: s
...
+        # Allow pinning back to current version. This case is specific to pinning because it
+        # is asynchronous, while starting is synchronous.
+        # If versions differ, continue to follow the logic in get_hup_action_type() below.
+        # The code here does not validate minimum OS version. The goal is to disable an earlier
+        # pin to a forward version; so the device already is running the "target" version.
+        if mode == "pin":
+            try:
+                # Must normalize first, since in format like "balenaOS 6.5.19".
+                parsed_current_ver = normalize_balena_semver(device_info["os_version"])
+                if semver_compare(parsed_current_ver, re.sub(".(dev|prod)$", "", target_os_version)) == 0:
```
kb2ma: @thgreasi , why does the node SDK use "\b" rather than "$" in the regex?

Also, why remove the check for .dev/.prod variants? Is the idea here just to cancel a previous pin for an upgrade?

balena/hup.py
```diff
@@ -48,6 +48,9 @@ def get_hup_action_type(device_type: str, current_version: Optional[str], target
     if Version.parse(target_version).compare(current_version) < 0:
         raise exceptions.OsUpdateError("OS downgrades are not allowed")
 
+    if Version.parse(target_version).compare(current_version) == 0:
+        raise exceptions.OsUpdateError("Current OS version matches Target OS version")
```
thgreasi, line 52: I would say that this can be an easy tiny patch PR of its own if you are up to it.

balena/models/device.py
thgreasi: I think I just copied that from balena-semver.
I guess I didn't think anything against it at that point b/c my goal was to jus remove any dev/prod variant and to not do any validity check, since
* balena-semver is doing the validation anyway
* that targetOsVersion comes directly from the DB's hostApp `release.raw_version` so it's expected to be valid.

I'm happy w/ your `$` & would approve though such a PR in the node sdk, but I also don't think it makes much difference to justify changing it atm.

thgreasi: You are right about the variant check missing. Here is a PR w/ the correct check:
https://github.com/balena-io/balena-sdk/pull/1611

balena/models/device.py
```diff
@@ -1760,6 +1780,54 @@ def start_os_update(
...
+            {
+                "$select": ["id", "uuid", "is_online", "os_version", "os_variant"],
+                "$expand": {"is_of__device_type": {"$select": "slug"}},
+            },
+        )
+
+        self.__check_os_update_target(device, target_os_version, "pin")
+
+        # In contrast to node SDK, includes only finalized versions.
+        all_versions = self.__device_os.get_available_os_versions(device["is_of__device_type"][0]["slug"])
```
thgreasi: Let's just call it `availableVersions`, since `all` as a term is generally used when we also include invalidated & draft OS releases.
```suggestion
        availableVersions = self.__device_os.get_available_os_versions(device["is_of__device_type"][0]["slug"])
```
Update: Oh I see we use the same var name in the start_os_update :disappointed: 
So, you can either leave this as is, or change both.

thgreasi (commented):
LGTM, just pending the variant check

balena/models/device.py
kb2ma: I agree it's more descriptive.

balena/models/device.py
kb2ma: Updated to use a similar check as the update to balena-sdk.

balena/hup.py
kb2ma: See #393

balena/models/device.py
```diff
@@ -228,6 +233,28 @@ def __check_os_update_target(self, device_info: TypeDevice, target_os_version: s
...
+            try:
+                # Must normalize first, since in format like "balenaOS 6.5.19".
+                parsed_current_ver = normalize_balena_semver(device_info["os_version"])
+                # Collect target version and variant.
+                res = re.match(r"(.+?)(?:\.(dev|prod))?$", target_os_version)
+                # Ensure version and variant match.
+                if (
+                    res.group(1)
+                    and semver_compare(parsed_current_ver, res.group(1)) == 0
+                    and device_info["os_variant"] == res.group(2)
```
thgreasi: Shouldn't this be like in the node sdk?
```suggestion
                    and (res.group(2) is None or device_info["os_variant"] == res.group(2))
```

balena/models/device.py
```diff
@@ -228,6 +233,28 @@ def __check_os_update_target(self, device_info: TypeDevice, target_os_version: s
...
+        if mode == "pin":
+            try:
+                # Must normalize first, since in format like "balenaOS 6.5.19".
+                parsed_current_ver = normalize_balena_semver(device_info["os_version"])
+                # Collect target version and variant.
+                res = re.match(r"(.+?)(?:\.(dev|prod))?$", target_os_version)
+                # Ensure version and variant match.
+                if (
+                    res.group(1)
+                    and semver_compare(parsed_current_ver, res.group(1)) == 0
```
thgreasi, line 251: NIT: I would move this check last in the `if` since it's expected to be the computationally most expensive.

balena/models/device.py
kb2ma: Fixed in [fa6abf4](https://github.com/balena-io/balena-sdk-python/pull/392/commits/fa6abf4aff88b43a27faa83d34a5427aa8875a34)

balena/models/device.py
kb2ma: Fixed in [fa6abf4](https://github.com/balena-io/balena-sdk-python/pull/392/commits/fa6abf4aff88b43a27faa83d34a5427aa8875a34)

kb2ma:
All comments addressed, @thgreasi .

thgreasi (approved):

PR #393, opened 2026-03-19
Prohibit OS update to same version via start_os_update()

kb2ma:
Bug fix. Prepares for #392 which does allow pin back to same version.

Change-type: patch


kb2ma:
Various Ubuntu build tests fail with each test run. I don't know the next step to resolve.

thgreasi (approved):

PR #394, opened 2026-03-20
Further increase log timeouts to resolve test failures

kb2ma:
We have consistently seen errors with logs tests in #392 and #393.

Experimenting with increase of log test timeouts to resolve.

Change-type: patch


kb2ma:
Timeout increase did not help.

PR #396, opened 2026-04-03
Disable log tests

kb2ma:
They run inconsistently; cause under investigation.

Change-type: patch


kb2ma:
Are we required to fix the SBOM failure first? What is the next step?

thgreasi (approved):

PR #401, opened 2026-06-09
Add team, team-application-access and team-membership models

JSReds:
See: https://balena.fibery.io/Work/Project/Add-Teams-Functionality-to-Python-SDK-2403

### DIFF
--- a/tests/functional/models/test_team.py
+++ b/tests/functional/models/test_team.py
@@ -97,6 +97,88 @@ class TestTeam(unittest.TestCase):
         self.assertNotIn(TestTeam.team1["id"], team_ids)
         self.assertNotIn(TestTeam.team2["id"], team_ids)
 
+    # --- TeamApplicationAccess ---
+
+    def test_14_application_access_get_all_by_team_empty(self):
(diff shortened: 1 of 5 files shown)


tests/functional/models/test_team.py
```diff
@@ -0,0 +1,115 @@
...
+        self.assertIn(TestTeam.team2["id"], team_ids)
+
+    def test_07_get(self):
+        # should return a specific team by id
+        team = self.balena.models.team.get(TestTeam.team1["id"])
+        self.assertEqual(team["id"], TestTeam.team1["id"])
+        self.assertEqual(team["name"], TestTeam.team1["name"])
+
+    def test_08_get_with_options(self):
+        # should support arbitrary pine options
```
otaviojacobi: NIT: for all these tests, can we remove the comment? Most of them the test name should be/already is already self explnatory

JSReds: Sure thing, I took that from `test_tag.py`, I thought it was something we wanted to make clearer

otaviojacobi (approved):

PR #402, opened 2026-06-18
Add request batching

JSReds:
See: https://balena.fibery.io/Work/Project/Add-batch-operation-handler-in-Python-SDK-2489

Change-type: minor

Description:

Introduce `batch_resource_operation_factory` to resolve, chunk, and validate lists of UUIDs/IDs before executing an operation (200 numeric / 50 string per chunk). Wire it into `device.move`, `pin_to_release`, `pin_to_supervisor_release`, and `application.remove`, which now all accept arrays in addition to a single uuid/id.

### DIFF
--- a/balena/models/application.py
+++ b/balena/models/application.py
@@ -24,6 +24,7 @@ from ..utils import (
     merge,
     with_supervisor_locked_error,
 )
+from ..request_batching import batch_resource_operation_factory
 from .device_type import DeviceType
 from .organization import Organization
 
@@ -47,6 +48,12 @@ class Application:
         self.build_var = BuildEnvVariable(pine, self)
         self.membership = ApplicationMembership(pine, self)
         self.invite = ApplicationInvite(pine, self, settings)
+
+        self.__batch_application_operation = batch_resource_operation_factory(
+            get_all=self.get_all,
+            not_found_error=exceptions.ApplicationNotFound,
+            ambiguous_error=exceptions.AmbiguousApplication,
+        )
 
     def __get_access_filter(self):
         return {
@@ -481,13 +488,13 @@ class Application:
 
         return self.__pine.post({"resource": "application", "body": body})
 
-    # TODO: enable batch operations
-    def remove(self, slug_or_uuid_or_id: Union[str, int]) -> None:
+    def remove(self, slug_or_uuid_or_id_or_ids: Union[str, int, List[int]]) -> None:
         """
         Remove application.
 
         Args:
-            slug_or_uuid_or_id (Union[str, int]): application slug (string), uuid (string) or id (number).
+            slug_or_uuid_or_id_or_ids (Union[str, int, List[int]]): application slug (string), uuid (string),
+                id (number) or ids (List[int]).
 
         Examples:
             >>> balena.models.application.remove('my_org/my_app')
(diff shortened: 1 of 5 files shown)


balena/utils/__init__.py
```diff
@@ -2,11 +2,11 @@
 import re
 from collections import defaultdict
 from typing import Any, Callable, Dict, Literal, Optional, TypeVar
-from .types.models import TypeDevice, TypeDeviceWithServices
+from ..types.models import TypeDevice, TypeDeviceWithServices
```
otaviojacobi: @JSReds why is this file move necessary?

JSReds: @otaviojacobi my thinking here was that `request_batching.py` needs to live alongside utils and import functionality from it. As far as I understand in python the cleanest way to support that is usually to turn `utils.py` into a package `utils/`, allowing both files to sit under the same namespace and letting `request_batching.py` use a straightforward relative import.

But here I invoke your wisdom and expertise, since I'm definitely not a Python expert 😅

otaviojacobi: @JSReds In general, I would try to avoid putting code inside `__init__.py`  as much as possible https://medium.com/@kay.herklotz/to-code-or-not-to-code-in-init-py-weighing-the-pros-and-cons-a35f158d388a

Given enough time and code it becomes annoying to mantain those as you start to see circular imports of modules and a few others the article above mention  :) I think you are fine to create request_batching.py outside of utils - it has clear specific function (request_batching) and does not need to live with utils just because the node-sdk does it this way

JSReds: @otaviojacobi thanks for the article! 🙏 

My point wasn't really to add code inside `__init__.py`, but simply to group all the utility related files into a single folder. In fact, the only change I made to `__init__.py` was updating the imports after moving the file into the new folder, I didn't add any new logic there.
That said, I understand your point that `request_batching.py` has a clear, specific purpose and doesn't necessarily belong in a utils package just because I am use to it :)

I've reverted the change and removed the utils folder I added in separated commit, if you are ok with it I can squash and rebase

balena/models/application.py
```diff
@@ -481,27 +488,35 @@ def create(
...
 
-    # TODO: enable batch operations
-    def remove(self, slug_or_uuid_or_id: Union[str, int]) -> None:
+    def remove(self, slug_or_uuid_or_id: Union[str, int, List[int]]) -> None:
         """
         Remove application.
 
         Args:
-            slug_or_uuid_or_id (Union[str, int]): application slug (string), uuid (string) or id (number).
+            slug_or_uuid_or_id (Union[str, int, List[int]]): application slug (string), uuid (string),
```
otaviojacobi: slug_or_uuid_or_id_or_ids? 
It is very long but at least it is consistent with the other ones....

balena/models/application.py
```diff
@@ -481,27 +488,35 @@ def create(
...
+                id (number) or ids (List[int]).
 
         Examples:
             >>> balena.models.application.remove('my_org/my_app')
             >>> balena.models.application.remove('c184556293854781aea71b0bdae10e45')
             >>> balena.models.application.remove(123)
         """
 
-        try:
+        if isinstance(slug_or_uuid_or_id, str):
```
otaviojacobi: I think this is a bug isn't it? This path should also be taken if `slug_or_uuid_or_id` is a single id (number), otherwise this will "work" but it gets compiled into a $in filter with a single element, which at SQL levl is less efficient than indexing directly by its id (it becomes a `= ANY()` with a binding param so it creates a full list and maybe be harder for the api to cache aswell)

otaviojacobi: More over, have you checked that ApplicationNotFound is still raised properly? I suspect yes, but it would be good to check if the tests cover it

balena/models/device.py
```diff
@@ -162,38 +169,18 @@ def __set(
...
+        def _operation(devices: list) -> None:
+            query: Any = {
+                "resource": "device",
+                "options": {"$filter": {"id": {"$in": [d["id"] for d in devices]}}},
+            }
+            if body is not None:
+                query["body"] = body
+            actual_fn(query)
+
+        self.__batch_device_operation(uuid_or_id_or_ids, _operation)
```
otaviojacobi: I think same comment as above? Ideally, if we get just a single uuid or single id, we prefer to use the resource_id indexing (e.g. directly passing an ID or {"uuid"...}, is the node-sdk also generalising over batch operations even when a single one is passing? 🤔

balena/models/device.py
```diff
@@ -667,56 +669,66 @@ def unset_custom_location(self, uuid_or_id_or_ids: Union[str, int, List[int]]) -
...
+            app_slug_or_uuid_or_id,
+            {
+                "$select": "id",
+                "$expand": {
+                    "is_for__device_type": {
+                        "$select": "is_of__cpu_architecture",
+                        "$expand": {"is_of__cpu_architecture": {"$select": "slug"}},
+                    }
+                },
             },
```
JSReds, line 698: I usually prefer to not create a variable if the query is small but if you prefer to keep the options separated I can revert it without issues

otaviojacobi (approved):

